PR #25, opened 2015-07-22
initialization of immutable objects

frankowskid:
1. Currently it is not possible to create objects if default constructor does not exist or is not public.
Situation where default constructor does not exist is common when implementing immutable classes.

2. When we have immutable object, there is no settes for object. This also cause that we are unable to create object(PropertyUtils throws exceptions as it uses setters)

To solve#1 - I'm using Objenesis which is able to create object even if public constructor is not available

To solve#1 - I'm using reflection, but befoer doing it I'm ensuring that field is accessible (public). After setting I'm restoring previous state

### DIFF
--- a/src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java
+++ b/src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java
@@ -27,8 +27,9 @@ package io.github.benas.jpopulator.impl;
 import io.github.benas.jpopulator.api.Exclude;
 import io.github.benas.jpopulator.api.Populator;
 import io.github.benas.jpopulator.api.Randomizer;
-import org.apache.commons.beanutils.PropertyUtils;
 import org.apache.commons.math3.random.RandomDataGenerator;
+import org.springframework.objenesis.Objenesis;
+import org.springframework.objenesis.ObjenesisStd;
 
 import java.lang.reflect.Array;
 import java.lang.reflect.Field;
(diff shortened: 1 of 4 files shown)


Toilal:
Thanks for this pull request !
I think it would be better to use the official [objenesis](https://github.com/easymock/objenesis) instead of the one repacked in Spring. It was just updated 12 hours ago with release 2.2.

```xml
<dependency>
	<groupId>org.objenesis</groupId>
	<artifactId>objenesis</artifactId>
	<version>2.2</version>
</dependency>
```

frankowskid:
Hi,
Sure, I will update my pull request as soon as the 2.2 version is available in public mvn repo.
I have read spring docs and you are right we should use orginal objenesis as one repacked in spring is "for internal use only".

src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java
```diff
@@ -27,8 +27,9 @@
 import io.github.benas.jpopulator.api.Exclude;
 import io.github.benas.jpopulator.api.Populator;
 import io.github.benas.jpopulator.api.Randomizer;
-import org.apache.commons.beanutils.PropertyUtils;
 import org.apache.commons.math3.random.RandomDataGenerator;
+import org.springframework.objenesis.Objenesis;
```
benas: Since we are using the official objenesis dependency, this should be `org.objenesis.Objenesis;` right? The same for `org.objenesis.ObjenesisStd;`

benas:
Hi @frankowskid ,

Great addition indeed :+1: 

jPopulator was designed to populate standard java beans (default constructor + getters and setters). But this restriction can be removed to support immutable classes as well.

I'll merge your PR request asap. Please see [my comment](https://github.com/benas/jPopulator/pull/25/files#r36045468) on the committed file.

And thank you @Toilal for suggesting to use the official dependency of objenesis.

Regards
Mahmoud

src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java
frankowskid: Yes, I missed import update after changing pom dependency. Thanks for an update.

PR #32, opened 2015-08-03
Maven multi modules

Toilal:
I think it's ready to merge.

This is a big change, I think you could consider releasing this as the next major release, even if API is still compatible with current codebase.

- All features have been splitted into maven modules.

- Every piece of random data is now generated through the [Randomizer](https://github.com/Toilal/jPopulator/blob/maven-multi-modules/jpopulator-api/src/main/java/io/github/benas/jpopulator/api/Randomizer.java) interface. It ensure consistency between randomizers used internally by jPopulator and those writen by the user.

- Internal randomizers are now declared in [RandomizerRegistry](https://github.com/Toilal/jPopulator/blob/maven-multi-modules/jpopulator-api/src/main/java/io/github/benas/jpopulator/api/RandomizerRegistry.java) implementations. Each registry is a set of Randomizers, and is automatically registered in PopulatorBuilder using Java Provider API ([ServiceLoader](https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html) + `META-INF/services/io.github.benas.jpopulator.api.RandomizerRegistry` file). Currently, there's 3 registries : [InternalRandomizerRegistry](https://github.com/Toilal/jPopulator/blob/maven-multi-modules/jpopulator-core/src/main/java/io/github/benas/jpopulator/randomizers/registry/InternalRandomizerRegistry.java), [JodaRandomizerRegistry](https://github.com/Toilal/jPopulator/blob/maven-multi-modules/jpopulator-joda-time/src/main/java/io/github/benas/jpopulator/randomizers/registry/JodaRandomizerRegistry.java) and [ValidationRandomizerRegistry](https://github.com/Toilal/jPopulator/blob/maven-multi-modules/jpopulator-validation/src/main/java/io/github/benas/jpopulator/randomizers/registry/ValidationRandomizerRegistry.java)

- [PopulatorBuilder](https://github.com/Toilal/jPopulator/blob/maven-multi-modules/jpopulator-core/src/main/java/io/github/benas/jpopulator/impl/PopulatorBuilder.java) now supports a `registerRandomizerRegistry` method to add custom randomizer registries. Also, `disableDefaultRegistries` allow disabling of default registries provided by Java Provider API.

- To skip a value, a [Randomizer](https://github.com/Toilal/jPopulator/blob/maven-multi-modules/jpopulator-api/src/main/java/io/github/benas/jpopulator/api/Randomizer.java) should now throw [RandomizerSkipException](https://github.com/Toilal/jPopulator/blob/maven-multi-modules/jpopulator-api/src/main/java/io/github/benas/jpopulator/api/RandomizerSkipException.java) instead of returning `null`. As a consequence, a Randomizer returning `null` will be effective and the `null` value will be set on the requested field. (required for @Null validation randomizer).

- An [EnumRandomizer](https://github.com/Toilal/jPopulator/blob/maven-multi-modules/jpopulator-api/src/main/java/io/github/benas/jpopulator/api/EnumRandomizer.java) interface has been added to the API to support future customization of enum random values.


Toilal:
To continue this work, I need a refactoring because some parts of the code is coupled and make some class moving note possible. 

Main problem is in [PopulatorImpl#populateSimpleType method](https://github.com/benas/jPopulator/blob/master/src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java#L181-L186) : architecture for custom randomizer is good and use one Randomizer implementation per type, but default supported types are implemented in two classes : [DefaultRandomizer](https://github.com/benas/jPopulator/blob/master/src/main/java/io/github/benas/jpopulator/impl/DefaultRandomizer.java) and [BeanValidationRandomizer](https://github.com/benas/jPopulator/blob/master/src/main/java/io/github/benas/jpopulator/impl/BeanValidationRandomizer.java).

Those two classes adds coupling on joda-time dependency for DefaultRandomizer, and javax-validation for BeanValidationRandomizer.

So in need to refactor those two classes using Randomizer implementations, and register them like custom randomizers in PopulatorImpl class when the jpopulator-joda-time/jpopulator-validation module is available.

benas:
I see the trick. That's fine, go for refactoring those classes. thx

Toilal:
@benas any about about this and #33 ?

benas:
Hello Rémi,

Sorry for being late, there are a lot of changes and I did not got time to check them all.

Everything looks good. I'll merge your PR as soon as I can.

Kind regards
Mahmoud

benas:
Hi Rémi, I finally got time to review your PR, sorry for being late.

> All features have been splitted into maven modules.

That's great!

> Every piece of random data is now generated through the Randomizer interface. It ensure consistency between randomizers used internally by jPopulator and those writen by the user.

That's fantastic!

> Internal randomizers are now declared in RandomizerRegistry implementations. Each registry is a set of Randomizers, and is automatically registered in PopulatorBuilder using Java Provider API (ServiceLoader + META-INF/services/io.github.benas.jpopulator.api.RandomizerRegistry file). Currently, there's 3 registries : InternalRandomizerRegistry, JodaRandomizerRegistry and ValidationRandomizerRegistry

J'achète :-)

> PopulatorBuilder now supports a registerRandomizerRegistry method to add custom randomizer registries.

* The concept of registry is a clean way to logically group randomizers together :+1: . But IMO this should not be a public API, the user can supply custom randomizers at construction time (which will be registered behind the scene in a UserRandomizersRegistry), but should not bother with registries.

* In regards to priorities, of course we can define it via annotation and so on, but this introduces a unnecessary complexity, I like to KISS.. I mean, to keep it simple :-). In reality, user defined randomizers may override default ones, that's it. There are only two priorities.

>  Also, disableDefaultRegistries allow disabling of default registries provided by Java Provider API.

With the new Service Loader architecture, disabling a default registry is as simple as not putting its jar in the classpath, which is the most elegant part pf the pattern. So this option is redundant.

> To skip a value, a Randomizer should now throw RandomizerSkipException instead of returning null. 

I'm not sure this is the right place to decide if the value should be skipped or not. Imagine I want to implement the Randomizer interface. When should I throw a ``RandomizerSkipException` ? I don't have information of the context in which the randomizer is used to decide if the value should be skipped or not. It is the caller of the randomizer that can decide if the value should be skipped or not (ie, the field is annotated with @Exclude or not). Do you see?
So Randomizer should simply generate random values for a given type.

> As a consequence, a Randomizer returning null will be effective and the null value will be set on the requested field. (required for @Null validation randomizer).

We can imagine a "NullRandomizer" in the `io.github.benas.jpopulator.randomizers.validation` (Null Object pattern) or simply set the field to null when the field is annotated with `@Null`.

> An EnumRandomizer interface has been added to the API to support future customization of enum random values.

That's fine, but do we really need a separate interface for that? Why not simply a `class EnumRandomizer implements Randomizer<E extends Enum>` that will generate a random value from the values set of the enum E?

Issue #24, opened 2015-04-08
recursion on populateBean, doesn't include excluded fields


DR9885:
Currently there is no way to ignore fields on sub objects. It would be better to either pass on the list, or support sub-fields, like "field.subfield". Either could work...

benas:
Hi Daniel,

Thank you for pointing out this issue. Actually there is a way to exclude a field in a sub-object using the `@Exclude` annotation. But in fact, there is no way to exclude sub-fields using the API.

As you suggest, the API should support the "field.subfield" syntax to exclude sub-fields. Great idea!
I'll plan that for next release. Keep tuned ;-)

Regards
Mahmoud

Toilal:
It would be also useful to exclude fields for every instanceof a given Class, regardless it's position in the instance graph and the name of the property.

Toilal:
I think there's an API design issue with excluded fields. It has been introduced in Populator interface, but I would have been better to define those exclusions in PopulatorBuilder class, and keep consistency with randomizer registration :

```java
public class PopulatorBuilder {
    ...
    // Existing method to register randomizer
    public PopulatorBuilder registerRandomizer(final Class type, final Class fieldType, final String fieldName, final Randomizer randomizer) {
        // ...
        return this;
    }

    // Register an ignore exactly the same way
    public PopulatorBuilder ignore(final Class type, final Class fieldType, final String fieldName) {
        // ...
        return this;
    }
    ...
}
```

```java
public interface Populator {
    // Excluded fields should be removed from populator
    <T> T populateBean(final Class<T> type);
    <T> List<T> populateBeans(final Class<T> type);
    <T> List<T> populateBeans(final Class<T> type, int size);
}
```

benas:
Hi @Toilal ,

> It would be also useful to exclude fields for every instanceof a given Class, regardless it's position in the instance graph

Sure, the API will support the syntax `field.subfield.subsubfield` regardless how deep the object graph is.

In regards to the second point:

> I would have been better to define those exclusions in PopulatorBuilder

I had also suggested to define exclusions at construction time in the Builder API. But as discussed in issue #9 , if we define exclusions at construction time, the populator instance would be unusable for types that do not have the excluded fields. Technically speaking, we can of course ignore them if they do not exist in the class but we would like to create the populator once and use it for every type as [said](https://github.com/benas/jPopulator/issues/9#issuecomment-62699107) by @prietopa .

Regards
Mahmoud

Toilal:
[...]

PR #33, opened 2015-08-12
Recursive support

Toilal:
This is based on #32, and should not be merged before #32.

- Adds support for any bean class that have recursion in it's structure. If a class has already been generated in the populate method call, it will reuse the instance already generated instead of trying to generate it again and again. (see tests)

- Allow to ignore fields deeply in the object graph using dotted notation, as talked in #24.


benas:
> If a class has already been generated in the populate method call, it will reuse the instance already generated instead of trying to generate it again and again.

That's cool, some sort of caching to improve performance.

> Allow to ignore fields deeply in the object graph using dotted notation, as talked in #24.

Many thanks for this, field exclusion will be easier this way.

PR #34, opened 2015-09-02
update PopulatorImpl to support classes that have a private default c…

feckertson:
Update PopulatorImpl to support classes having a private default constructor.

### DIFF
--- a/src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java
+++ b/src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java
@@ -33,6 +33,7 @@ import org.objenesis.ObjenesisStd;
 
 
 import java.lang.reflect.Array;
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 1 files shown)


benas:
Hello, thank you for your PR! jPopulator already supports private constructors through objenesis. Here is a passing test:

```java

public class BeanWithPrivateDefaultConstructor {

    private String name;
    
    private BeanWithPrivateDefaultConstructor() {
    }

    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }
}

    @Test
    public void testBeanWithPrivateDefaultConstructorPopulation() throws Exception {
        BeanWithPrivateDefaultConstructor bean = populator.populateBean(BeanWithPrivateDefaultConstructor.class);
        assertThat(bean).isNotNull();
        assertThat(bean.getName()).isNotNull();
    }
```

So I think there is no need to change the visibility of the constructor. Objenesis will do it [here](https://github.com/feckertson/jPopulator/blob/private-constructor/src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java#L108).
Do you see the trick?

feckertson:
Yes, I agree. The code in master covers this matter. 

However, the test passes if my pom references a version of jPopulate that I build myself, but not if reference 1.2.0. I have tried this with 32 and 64 bit windows versions of jdk1.7.0_79 using each of maven 3.2.1, 3.2.5 and 3.3.3. I duplicated the tests on my home device to ensure the problem has nothing to do with Cerner's mirror of maven central.

All I can figure is the end state that is deployed as 1.2.0 does not match the 1.2.0 code base. Please check out https://github.com/feckertson/jPopulatorConsumer and perform a maven test to confirm the problem.

PR #35, opened 2015-09-02
Random collections

feckertson:
Update PopulatorImpl to randomly populate collections rather than setting them equal to an empty collection. Also, changed the default maximum number of items added by populateBeans to be 100 not Short.MAX_VALUE

PR #38, opened 2015-09-12
Implement random collections including generic type handling

feckertson:
This PopulatorImpl populates collection attributes with random elements applying generic types as appropriate. 

Not all collection attributes can actually be handled generically. For example the add method of an ArrayBlockingQueue instance will fail unless the instance has been properly initialized which will not be the case because PopulatorImpl initializes classes randomly. This implementation will throw an exception when such cases are encountered which seems better than quietly populating an empty collection object. A custom populator can always be applied for now.

### DIFF
--- a/src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java
+++ b/src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java
@@ -24,16 +24,7 @@
 
 package io.github.benas.jpopulator.impl;
 
-import io.github.benas.jpopulator.api.Exclude;
-import io.github.benas.jpopulator.api.Populator;
-import io.github.benas.jpopulator.api.Randomizer;
-import org.apache.commons.math3.random.RandomDataGenerator;
-import org.objenesis.Objenesis;
-import org.objenesis.ObjenesisStd;
-
-
 import java.lang.reflect.Array;
-import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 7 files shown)


benas:
Hi, it looks great! Thank you.

Just a quick note, this PR includes unnecessary code to support classes with a private default constructor (as discussed in #35 ) Could you please remove it and keep only changes related populating random generic collections? 

I'll merge your changes asap.

Many thanks upfront.
Mahmoud

feckertson:
That code has already been removed.

In case you did not see it....
I added a comment on #34 suggesting that the 1.2.0 release might be broken. 

If I let maven pull 1.2.0 from maven central, private default constructors cause problems, but if I build 1.2.0 from source, they do not. https://github.com/feckertson/jPopulatorConsumer is a trivial example with a private default constructor to demonstrate (assuming there is not a already working build of 1.2.0 in the local m2 repo).

src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java
```diff
@@ -139,9 +164,24 @@ public PopulatorImpl() {
             throw new IllegalArgumentException("The number of beans to populate must be positive.");
         }
         List<Object> beans = new ArrayList<Object>();
-        for (int i = 0; i < size; i++) {
-            Object bean = populateBean(type, excludedFields);
-            beans.add(bean);
+        if (isSupportedType(type)) {
+            BeanFieldHolder beanFieldHolder = new BeanFieldHolder();
```
benas: I'm not sure I understand the purpose of the `BeanFieldHolder` class. What does this block do?

```java
BeanFieldHolder beanFieldHolder = new BeanFieldHolder();
Field field = null;
try {
   field = BeanFieldHolder.class.getField("field");
} catch (Exception e) {
    // this should never happen
    e.printStackTrace();
}
```

benas:
> That code has already been removed.

Thanks for the update!

> In case you did not see it..I added a comment on #34 suggesting that the 1.2.0 release might be broken.

I saw it, no problem. The master branch contains the code for current snapshot (for next release, not current release). You can import the snapshot version in your project until the next release is out, I'll publish the snapshot version asap.

> This implementation will throw an exception when such cases are encountered which seems better than quietly populating an empty collection object. 

Interesting point. Until now, since collection population was not supported, I decided to generate empty collections instead of throwing an exception. Now that this feature is here (thanks to you :+1: ), we can throw an exception for unsupported types.

I'll probably get conflicts when merging your PR with other active PRs, I'll fix them and keep you informed.

Many thanks for your contribution!
Kind regards
Mahmoud

src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java
feckertson: The code needs an instance of java.lang.reflect.Field of type Object to pass into populateSupportedType at line 177, and I don't know how to construct one. 

I agree the code is a bit sloppy here. 
- There is no need to instantiate beanFieldHolder which is never used
- A  constant, say OBJECT_FIELD, could be used in place of the local variable field.
- The name ObjectFieldSupplier would better match its use than BeanFieldHolder. 

How would you like me to proceed?

kermit-the-frog:
Any updates on this request?

benas:
Hi @feckertson, @kermit-the-frog 

No, this PR has conflicts with the current master branch and I didn't got time to fix them, sorry for that.

I'll try to do it asap and keep you informed.

Regards
Mahmoud

benas:
Hi,

I'm really sorry for my late reply, shame on me ..
Could you please update this PR with the current code in the master branch?
I'm planning to merge it for the next release.
Thank you in advance

Kind regards
Mahmoud

feckertson:
Closing this PR. I have opened a new one that is consistent with the current master branch.

PR #45, opened 2015-12-22
slight simplification of some tests

PascalSchumacher:
I found some test which can be simplified a bit.


benas:
Fantastic! Thank you.

I'm in love with AssertJ :smile:

PascalSchumacher:
Yes, AssertJ is great, even after over a year of using it I still discover new features all the time. :)

PR #47, opened 2016-01-03
Implement random collections including generic type handling with review comments addressed.

feckertson:


benas:
Great! Thank you for the update.

This was the last feature planned for the 2.0 release, I'll update the documentation and prepare the release.

Many thanks for this addition

Kind regards

benas:
Hello Fred,

Looks like we are not yet done with this feature:
* primitive arrays cause an exception (for example field of type `int[] numbers;`)
* composite collections cause an exception (for example `Map<String, List<String>> synonyms;`)
* Many collection types are not supported yet: BlockingQueue, HashTable, LinkedHashSet, ConcurrentHashMap, and many others.

I didn't imagined that populating a Java collection with random data would be that hard :rage: 
I will fix these issues in a future release, this was just for your information.

Regards
Mahmoud

feckertson:
I'm disappointed I didn't think to test that. It's pretty obvious that it can't work now that you point it out.

PR #48, opened 2016-01-07
simplification assertions in random collection tests

PascalSchumacher:
Code duplication could be reduced further by adding some helper methods for common assertions e.g. a method for `assertThat(Collection).isNotEmpty().hasOnlyElementsOfType(String.class).doesNotContain(null, "")`. If you think that is a good idea I can create another pull request for that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ If you have any question, suggestion, or feedback, do not hesitate to use the [G
 * [Nikola Milivojevic](https://github.com/dziga)
 * [Rémi Alvergnat](http://www.pragmasphere.com)
 * [Fred Eckertson](https://github.com/feckertson)
+* [Pascal Schumacher](https://github.com/PascalSchumacher)
 
 Thank you all for your contributions!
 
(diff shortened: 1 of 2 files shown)


benas:
Hello,

Great! That's great..

I'm of course against code duplication in production code, but for tests, some times it is easier to read inlined assertions (even if they are duplicated) than jumping between methods. But if you see that it would be better, go for it!

Please don't hesitate to add yourself to the contributors list in README.md.

Many thanks
Best regards
Mahmoud

PascalSchumacher:
Hi Mahmoud,

thanks. :)

>I'm of course against code duplication in production code, but for tests, some times it is easier to read 
>inlined assertions (even if they are duplicated) than jumping between methods.

I agree, it is important to keep test simple and readable. I also have problems finding good names for further custom assertion(methods), so lets keep the pull request as it is.

I have also added myself as a contributor.

Thanks,
Pascal

benas:
Thanks for the update!

PR #50, opened 2016-01-17
fix getRandomArray to handle primitives

feckertson:
Here is a fix. I tried but couldn't find a cleaner way to handle it.
I won't be able to spend any time on side projects for the next few weeks.

I really like the cleanup of the collection population tests, by the way.


benas:
Many thanks for the quick fix and for your effort on this feature.

PR #52, opened 2016-01-24
Some minimal improvements

PascalSchumacher:


PascalSchumacher:
I guess 01acb98ad7285585d5a194ff75519433df7a851d is a controversial, so feel free to reject if you disagree with this change. :smile:

benas:
That's fine, thx for the update.

Adding generics like the following change is great:

```diff
- Randomizer getRandomizer(final Class<?> type);
+ <T> Randomizer<T> getRandomizer(final Class<T> type);
```

but I'm not sure I understand the value of wildcards here:

```diff
- private final Map<Class, Randomizer> randomizers = new HashMap<Class, Randomizer>();
+ private final Map<Class<?>, Randomizer<?>> randomizers = new HashMap<>();
```

To be honest, I've never had to deal with nasty details between List<Object>, List<?> and raw List (as detailed [here](https://docs.oracle.com/javase/tutorial/java/generics/unboundedWildcards.html)) .. Java generics are full of tricky tricks :smile:

PascalSchumacher:
Imho there are two "values" of `<?>`:

1. You suppress the compiler warnings. 
2. `List <?>` makes it explicit that you do not know the type of the objects in this list. (Pre-Java-5 you had to write `List` even if you always expected `List<String>`.)

So it's not a lot, but adding the `<?>` is not a lot of work. :smile:

benas:
> You suppress the compiler warnings.

Even without the wildcard I don't have compiler warning (using Java 7):

 `private Map<Class, Randomizer> randomizers = new HashMap<>();`

So I think for a parametrized type like `Randomizer<T>`,  writing `Randomizer<?> randomizer` or `Randomizer randomizer` are equivalent when we don't know the type upfront, but I completely agree with you, it is more explicit with the wildcard than without it..  and as you said, it's not very expensive to add it :+1:

PascalSchumacher:
Depending on the IDE/tools you use you may not see the warnings, but they are there. If I execute `mvn compile`, I get:

```
...
[INFO] /random-beans/random-beans/src/main/java/io/github/benas/randombeans/PopulatorImpl.java: PopulatorImpl.java uses unchecked or unsafe operations.
[INFO] /random-beans/random-beans/src/main/java/io/github/benas/randombeans/PopulatorImpl.java: Recompile with -Xlint:unchecked f
or details.
...
```
If I add this:
```
<compilerArgs>
  <arg>-Xlint:all</arg>
</compilerArgs>
```
to the maven compiler plugin configuration I get:

```
...
[WARNING] /random-beans/random-beans/src/main/java/io/github/benas/randombeans/PopulatorImpl.java:[165,32] found raw type: java.lang.Class
  missing type arguments for generic class java.lang.Class<T>
...
```

Eclipse shows these kind of warnings by default, so I like to get rid of them. :smile:

benas:
I revert my last comment :smile: Sorry for that. `-Xlint:all` should be the default !

That said, there are many warnings and the code is a bit messy. Now that all features for v2 are implemented and tests are green, time for refactoring and cleaning the code.. Make it work than make it better :wink:

PascalSchumacher:
That's also my favorite approach. :smile:

PR #53, opened 2016-01-24
update assertj to 2.3.0

PascalSchumacher:


benas:
Thanks, indeed this is the last version of assertj compatible with Java 7 (Random Beans v2).

I'll upgrade to last version of assertj for Random Beans v3 which will requires Java 8

PR #55, opened 2016-01-26
minimal performance improvement: create ArrayLists with size paramete…

PascalSchumacher:
…r, if size is known in advance

I guess in context of random-beans this does not really matter, but it wasn't much work, so why not?


benas:
Hi Pascal,

Indeed, it's better to create the array with the specified size upfront, nice spot.
It's a pleasure to have contributors like you :smile:  Thank you for all these additions!

Kind Regards
Mahmoud

PR #59, opened 2016-01-31
Add ability to populate a interface and abstract cla…

PascalSchumacher:
…ss field types with a random concrete subtype

I could not use random-beans to populate an object because it contained fields of the abstract class XMLGregorianCalendar. So I hacked together a quick prototype to see how difficult it is to add something like this to random-beans.

Obviously this is far from finished, but it should be enough to judge if this would be a worthy addition to random-beans?  No problem if you think this is out of scope :smile:.

### DIFF
--- a/random-beans/src/main/java/io/github/benas/randombeans/PopulatorImpl.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/PopulatorImpl.java
@@ -31,14 +31,14 @@ import io.github.benas.randombeans.api.Populator;
 import io.github.benas.randombeans.api.Randomizer;
 import io.github.benas.randombeans.api.RandomizerRegistry;
 import io.github.benas.randombeans.randomizers.EnumRandomizer;
-import io.github.benas.randombeans.util.ReflectionUtils;
 import io.github.benas.randombeans.randomizers.SkipRandomizer;
+
+import org.apache.commons.lang3.RandomUtils;
 import org.objenesis.Objenesis;
 import org.objenesis.ObjenesisStd;
 
 import java.lang.reflect.Field;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 
@@ -122,11 +122,22 @@ final class PopulatorImpl implements Populator {
         }
     }
 
-    private <T> Class<?> randomConcreteSubTypeOf(final Class<T> type) throws ClassNotFoundException {
-        List<Class<?>> concreteSubtypes = ReflectionUtils.getPublicConcreteSubTypesOf(type);
-        if (concreteSubtypes.isEmpty()) throw new ClassNotFoundException("Unable to find a concrete subtype of type: " + type);
-        Collections.shuffle(concreteSubtypes);
-        return concreteSubtypes.get(0);
+    private Class<?> randomConcreteSubTypeOf(final Field field) throws ClassNotFoundException {
+        Class<?> fieldType = field.getType();
+        List<Class<?>> concreteSubTypes = getPublicConcreteSubTypesOf(fieldType);
+        concreteSubTypes = findSubTypesWithSameParameterizedTypes(field, concreteSubTypes);
+        if (concreteSubTypes.isEmpty()) throw new ClassNotFoundException("Unable to find a matching concrete subtype of type: " + fieldType);
+        return randomElementOf(concreteSubTypes);
+    }
+
+    private Class<?> randomConcreteSubTypeOf(final Class<?> type) throws ClassNotFoundException {
+        List<Class<?>> concreteSubTypes = getPublicConcreteSubTypesOf(type);
+        if (concreteSubTypes.isEmpty()) throw new ClassNotFoundException("Unable to find a concrete subtype of type: " + type);
+        return randomElementOf(concreteSubTypes);
+    }
+
+    private Class<?> randomElementOf(List<Class<?>> concreteSubTypes) {
+        return concreteSubTypes.get(RandomUtils.nextInt(0, concreteSubTypes.size()));
     }
 
     @Override
@@ -153,7 +164,7 @@ final class PopulatorImpl implements Populator {
     }
 
     private void populateField(final Object target, final Field field, final PopulatorContext context)
-            throws BeanPopulationException, IllegalAccessException {
+            throws BeanPopulationException, IllegalAccessException, ClassNotFoundException {
 
         Randomizer<?> randomizer = randomizerProvider.getRandomizerByField(field);
         if (randomizer instanceof SkipRandomizer) {
@@ -171,7 +182,7 @@ final class PopulatorImpl implements Populator {
     }
 
     private Object generateRandomValue(final Field field, final PopulatorContext context)
-            throws BeanPopulationException, IllegalAccessException {
+            throws BeanPopulationException, IllegalAccessException, ClassNotFoundException {
         Class<?> fieldType = field.getType();
         Object value;
         if (isArrayType(fieldType)) {
@@ -181,7 +192,12 @@ final class PopulatorImpl implements Populator {
         } else if (isMapType(fieldType)) {
             value = mapPopulator.getRandomMap(field);
         } else {
-            value = doPopulateBean(fieldType, context);
+            if (scanClasspathForConcreteClasses && isAbstract(fieldType)) {
+                // create a new instance of a random concrete subtype
+                value = objenesis.newInstance(randomConcreteSubTypeOf(field));
+            } else {
+                value = doPopulateBean(fieldType, context);
+            }
         }
         return value;
     }
--- a/random-beans/src/main/java/io/github/benas/randombeans/PopulatorImpl.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/PopulatorImpl.java
@@ -31,12 +31,14 @@ import io.github.benas.randombeans.api.Populator;
 import io.github.benas.randombeans.api.Randomizer;
 import io.github.benas.randombeans.api.RandomizerRegistry;
 import io.github.benas.randombeans.randomizers.EnumRandomizer;
+import io.github.benas.randombeans.util.ReflectionUtils;
 import io.github.benas.randombeans.randomizers.SkipRandomizer;
 import org.objenesis.Objenesis;
 import org.objenesis.ObjenesisStd;
 
 import java.lang.reflect.Field;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 
@@ -60,6 +62,8 @@ final class PopulatorImpl implements Populator {
     private MapPopulator mapPopulator;
 
     private RandomizerProvider randomizerProvider;
+    
+    private boolean scanClasspathForConcreteClasses;
 
     PopulatorImpl(final Set<RandomizerRegistry> registries) {
         objenesis = new ObjenesisStd();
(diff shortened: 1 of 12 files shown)


benas:
Hi,

> how difficult it is to add something like this to random-beans.

Supporting new types is easy if you use `RandomizerRegistry`. This abstraction associates a type to a given randomizer that will be used to generate random values when a field of this type is encountered in the object graph.

Randomizer registries are discovered through the SPI, so you can create a module containing an implementation declared in `META-INF` and you are done. Or you can register custom registries through the `PopulatorBuilder` API.

For example, supporting Joda time types is done through a [custom randomizer registry](https://github.com/benas/random-beans/blob/master/random-beans-jodatime/src/main/java/io/github/benas/randombeans/randomizers/jodatime/registry/JodaTimeRandomizerRegistry.java). Random Beans also uses [registries internally](https://github.com/benas/random-beans/tree/master/random-beans/src/main/java/io/github/benas/randombeans/randomizers/registry) to support built-in Java types and annotated fields.

You can add support for `XMLGregorianCalendar` using a custom registry.

Hope this helps

PascalSchumacher:
Hi Mahmoud,

thanks for the detailed explanation. :smile: 

I know this is possible and not that hard, but it still is a bit of work even for a single custom type. 

The aim of this pull request is to enable random-beans to populate classes with fields of interface and abstract class types out of the box, without requiring any user configuration.

benas:
Hello Pascal,

> The aim of this pull request is to enable random-beans to populate classes with fields of interface and abstract class types out of the box

I just looked quickly at the changes (without testing) trying to understand. So roughly speaking, the idea is to pick up a random implementation from the classpath for each field of type interface or abstract class. Is this correct? Please correct me if I'm wrong. So it's a general purpose support, not for a specific type like we would do with registries.

If your field is of type `XMLGregorianCalendar`, you still need an implementation of this abstract class in the classpath right? 

> it still is a bit of work even for a single custom type.

I agree, the idea is not to create a registry per type. Just like joda time types or Java 8 types, there is a registry per a whole set of types.

> without requiring any user configuration.

If Random Beans provides a module to support the `XMLGregorianCalendar` (and others abstract types), I'm not sure there would be a configuration required, you just need to add the module to your classpath and you get support for custom types, this is the most elegant part of the service locator pattern. I mentioned the possibility to use the API but this is optional.

Regards
Mahmoud

PascalSchumacher:
>I just looked quickly at the changes (without testing) trying to understand. So roughly speaking, the idea is to pick up a random implementation from the classpath for each field of type interface or abstract class. Is this correct? Please correct me if I'm wrong. So it's a general purpose support, not for a specific type like we would do with registries

correct

>If your field is of type XMLGregorianCalendar, you still need an implementation of this abstract class in the classpath right?

correct

>I agree, the idea is not to create a registry per type. Just like joda time types or Java 8 types, there is a registry per a whole set of types.

Sorry, I did not express myself clearly. Of course you can use a single registry, but you still need to write a randomizer for every interface and abstract class I guess?

>If Random Beans provides a module to support the XMLGregorianCalendar (and others abstract types), I'm not sure there would be a configuration required, you just need to add the module to your classpath and you get support for custom types, this is the most elegant part of the service locator pattern. I mentioned the possibility to use the API but this is optional.

Yes, there would be no configuration required. It's an elegant solution. :smile: But it would be too much work to write a randomizer even for every interface / abstract class in the JDK, to speak nothing of libraries and custom classes.

benas:
> but you still need to write a randomizer for every interface and abstract class I guess?
> it would be too much work to write a randomizer even for every interface / abstract class in the JDK, to speak nothing of libraries and custom classes.

Yes yes of course now that I understand the purpose of your PR I see the point. Your addition is a general purpose solution that would work with any interface or abstract type by scanning the classpath, not by implementing a randomizer for every single type. Nice one :+1: 

I think we can make this configurable at construction time, like for example:

```java
Populator p = PopulatorBuilder.aNewPopulatorBuilder()
   .scanClasspath(true) // or other method name
   .build();
```

What do you think?

PascalSchumacher:
Yes, I believe making it optional is good idea, because scanning the class path is a bit slow and can have surprising/confusing results.

PascalSchumacher:
Rebased and extended.

I made it completely optional. If a user does not wish to use the features he will not need the dependencies, because the ReflectionsUtils class will never be loaded.

Feel free to suggest any changes.

benas:
Fantastic! Thank you for this addition.

I really like your code, clean, clear and easy :+1: 

Just a couple of notes, can you please:

* Move methods of the `ReflectionsUtils` class in the existent `ReflectionUtils` class? This would remove a duplicate class for reflection utilities
* Since the `scanClasspathForConcreteClasses` parameter is optional, you can use a setter in the `PopulatorImpl` to set its value. I prefer using constructor parameters for mandatory fields and setters for optional fields. Do you agree?

Everything else is perfect :smile:

random-beans/src/main/java/io/github/benas/randombeans/PopulatorImpl.java
```diff
@@ -108,11 +118,18 @@
...
             return result;
-        } catch (Exception e) {
+        } catch (InstantiationError | Exception e) {
             throw new BeanPopulationException("Unable to generate a random instance of type " + type, e);
         }
     }
 
+    private <T> Class<?> randomConcreteSubTypeOf(final Class<T> type) {
+        List<Class<?>> concreteSubtypes = ReflectionsUtils.getConcreteSubTypesOf(type);
+        if (concreteSubtypes.isEmpty()) throw new NullPointerException("Unable to find a concrete sub type of type " + type);
```
benas: here, I think we can simply warn the user and silently ignore the field instead of making the whole population fail with NPE. What do you think?

random-beans/src/test/java/io/github/benas/randombeans/PopulatorImplTest.java
```diff
@@ -145,6 +146,23 @@ public void javaNetTypesShouldBePopulated() throws Exception {
...
+    @Test
+    public void generatesConcreteTypesForInterfacesAndAbstractClassesIfScanClasspathForConcreteClassesIsEnabled() throws Exception {
+        Populator scanningPopulator = aNewPopulatorBuilder().scanClasspathForConcreteClasses(true).build();
+
+        Mamals mamals = scanningPopulator.populateBean(Mamals.class);
+
+        assertThat(mamals.getMamal()).isOfAnyClassIn(Human.class, Ape.class, Person.class, SocialPerson.class);
+        assertThat(mamals.getMamalImpl()).isOfAnyClassIn(Human.class, Ape.class, Person.class, SocialPerson.class);
+    }
+
```
benas: I'm curious about how the new code will behave if I have a field of type `Comparable` in my bean:

```java
class Person {
   private Comparable c1;
   private Comparable<Date> c2;
}
```

There are hundreds of comparable implementations out there (635 in the classpath of the project) and I would like to know how the code will behave if the implementation that has been chosen randomly is abstract, like `sun.org.mozilla.javascript.internal.ast.AstNode` or `com.sun.tools.javadoc.ProgramElementDocImpl`.. Do you see the trick? Can you please add a test case with a `Comparable` field? Many thanks

PascalSchumacher:
>Move methods of the ReflectionsUtils class in the existent ReflectionUtils class? This would remove a >duplicate class for reflection utilities

Sure. I moved the code out of ReflectionUtils to make the dependency on the reflections library (and its dependencies guava and java-assist) optional. If it's o.k. to add reflections as a non-optional dependency I will move the code back. 

>Since the scanClasspathForConcreteClasses parameter is optional, you can use a setter in the >PopulatorImpl to set its value. I prefer using constructor parameters for mandatory fields and setters for >optional fields. Do you agree?

Sure. I thought about making it a setter, but I was not sure which style you prefer. I will update the code.

random-beans/src/test/java/io/github/benas/randombeans/PopulatorImplTest.java
PascalSchumacher: Sure. Will do tomorrow.

benas:
Great thanks.

> If it's o.k. to add reflections as a non-optional dependency I will move the code back.

In regards to reflections library, we can add it to dependencies, no problem. If with this library we can remove the `ReflectionUtils` class, this would be better! Good coders code, great coders reuse :wink: 

I didn't knew about it, I will take a look and try to reuse a maximum of its utilities in Random Beans.

random-beans/src/test/java/io/github/benas/randombeans/PopulatorImplTest.java
PascalSchumacher: Nice catch :), at the moment this fails with:

>Caused by: io.github.benas.randombeans.api.BeanPopulationException: Unable to generate a random instance of type class org.apache.commons.lang3.builder.DiffBuilder
>Caused by: java.lang.ClassCastException: sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl cannot be cast to java.lang.Class

I will try add support for this.

PascalSchumacher: According to http://tutorials.jenkov.com/java-reflection/generics.html you need the field to determine the generic type of the `Comparable`. Sadly in `doPopulateBean()` you do not have access to the field type anymore. So it is not possible choose a subclasses which implements  `Comparable<Date>`. The current implementation just chooses a random `Comparable`.

PascalSchumacher:
Updated the pull request.

PascalSchumacher:
I added another commit which solves the `Comparable<Date>` problem for fields. Now only sub types with the same generic type parameters are used to populate a field.

PascalSchumacher:
rebased to latest master

benas:
Thank you for the update!

gpaul-idexx:
The static initialization of the ReflectionUtils.reflections field takes a few seconds on my machine. That's a pretty big loss in performance. It competes against my interest in keeping unit tests fast; particularly noticeable when I just want to run a single test in my IDE and the wait is pretty lengthy.

PascalSchumacher:
There is an easy solution for this. We just need to move the ReflectionUtils stuff back to it's now class (like in my first version of the pull request), then the static initialization is only done when necessary.

benas:
Hi @gpaul-idexx 
We will try make this dependency optional in case you don't activate class path scanning for concrete types.

@PascalSchumacher Thank you for your quick reply. Can you please make the changes as in your first PR? Many thanks upfront

Best regards
Mahmoud

PascalSchumacher:
@benas Np, here's the pull request https://github.com/benas/random-beans/pull/79 You can use the jvm arg "-verbose:class" (which displays classes as the are loaded) to check that the classes are not loaded.

benas:
@PascalSchumacher Thank you for the quick update

@gpaul-idexx We have made the dependency to Reflection library optional. Now, if you don't activate the `scanClasspathForConcreteTypes` parameter, the static bloc should not be executed.

I've deployed version 3.0.0-SNAPSHOT to maven central, can you give it a try and tell us if it is faster with this version? Thank you in advance

gpaul-idexx:
That change does fix the slowness. I didn't see the new version in maven central (yet), so I just built a copy from master and tried it.

Thank you

benas:
> That change does fix the slowness

Good news! Thank you for taking time to test.

gpaul-idexx:
Do you think you'll be releasing this patch?

Alternatively, can you confirm that the snapshot is available on maven? (I'd prefer not to set up a dependency on a local build that includes this fix).

Thanks again

benas:
Hi,

> Do you think you'll be releasing this patch

The release date of v3.0.0 is not decided yet

> Alternatively, can you confirm that the snapshot is available on maven

Yes, it is online: https://oss.sonatype.org/content/repositories/snapshots/io/github/benas/random-beans/3.0.0-SNAPSHOT/

> I'd prefer not to set up a dependency on a local build that includes this fix

No need to setup a local build, you need to declare the sonatype snapshot repository in your pom to be able to import it in your project, see [here](https://github.com/benas/random-beans/wiki/Getting-Started#maven-usage).

Kind regards
Mahmoud

PR #70, opened 2016-02-08
Mimimal cleanup

PascalSchumacher:


benas:
Thank you for these improvements. Usually, this is what I do when I prepare a release: checking licence headers, pretty print, update javadoc. I also run a sonar analysis to check if I have some major issues..

Now that v2.0 is feature complete, I'll prepare the release for the next weekend (I hope :smile: ).

Issue #66, opened 2016-02-07
Should collections without generics be generated empty?


PascalSchumacher:
At the moment collections without generic types are filled `String`s. I wonder if it would be more generally useful if they were just empty?

benas:
> At the moment collections without generic types are filled Strings.

Yes, this has been added in [this PR](https://github.com/benas/random-beans/pull/38/files#diff-ded1e47d146cf5f4d680ed6e256f966bR313). The choice of String as a base class for objects in raw collections has been made because some collections require Comparable objects (see comment of the author of the PR)

> I wonder if it would be more generally useful if they were just empty?

As of version 1.x, Random Beans generates empty collections. Generating non-empty collections has been [requested](https://github.com/benas/random-beans/issues/19) and planned for release 2.x. What makes you think that it would be more useful is they were just empty?

PascalSchumacher:
I agree that generating non-empty collections is useful :+1: , but only if the collection contains elements of the right type.

Putting strings into collections without generics is guesswork. It's just a gut feeling, but I fear that this will be the wrong a lot.

BTW: Random Beans 2.0 can not generate Collections with generic place holders: e.g. `List<?>`, `List<? extends Runnable>` etc. fail, because Random Beans tries to create beans for `?`, `? extends Runnable` etc. Maybe in these cases an empty collection should be generated instead?

benas:
Hello Pascal,

You are right and I think that we think the same way :smile: See my comment on the subject here: https://github.com/benas/random-beans/issues/19#issuecomment-71988275 : 

> if the field is a non-typed collection (private Set accounts), then jPopulator cannot know which type of objects to generate. Hence, it will generate an empty collection (as it does now).

When I got the PR, I said "hmm why not.." and then it was merged, but may be too quickly. As you said, Random Beans should not make assumptions or guess what type it should fill in the collection, so if there are people in 2016 still using raw collections, then they should accept empty collections or use a custom randomizer for the field.

That said, I'll revert back the code to generate empty collections for raw collections..

Thank you for pointing this out just before the release!

Best regards
Mahmoud

PascalSchumacher:
:smile:

benas:
That's done. Raw collection are generated empty now. The code is even simpler!

[...]

PR #71, opened 2016-02-09
Add ability to populate beans with collections with unbounded wildcar…

PascalSchumacher:
…ds. These collections are generated as empty collections.

Before random-beans would fail with:

>Caused by: java.lang.ClassCastException: sun.reflect.generics.reflectiveObjects.WildcardTypeImpl cannot be cast to java.lang.Class
	at io.github.benas.randombeans.CollectionPopulator.getRandomCollection(CollectionPopulator.java:53)
	at io.github.benas.randombeans.BeanPopulator.generateRandomValue(BeanPopulator.java:176)
	at io.github.benas.randombeans.BeanPopulator.populateField(BeanPopulator.java:164)
	at io.github.benas.randombeans.BeanPopulator.populateFields(BeanPopulator.java:149)
	at io.github.benas.randombeans.BeanPopulator.doPopulateBean(BeanPopulator.java:123)
	... 26 more

### DIFF
--- a/random-beans/src/main/java/io/github/benas/randombeans/util/ReflectionUtils.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/util/ReflectionUtils.java
@@ -33,6 +33,7 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.lang.reflect.WildcardType;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -162,6 +163,24 @@ public abstract class ReflectionUtils {
      */
     public static boolean isParameterizedType(final Type type) {
         return type != null && type instanceof ParameterizedType && ((ParameterizedType) type).getActualTypeArguments().length > 0;
+    }
+
+    /**
+     * Check if a type is an unbounded wildcard type
+     * @param parameterizedType the ParameterizedType to check
+     * @return true if the ParameterizedType is an unbounded wild card type, false otherwise
+     */
+    public static boolean isUnboundedWildcardType(final ParameterizedType parameterizedType) {
+        Type[] actualTypeArguments = parameterizedType.getActualTypeArguments();
+        Type type = actualTypeArguments[0];
+        if (actualTypeArguments.length == 1 && type instanceof WildcardType) {
+            WildcardType wildcardType = (WildcardType) type;
+            Type upperBound = wildcardType.getUpperBounds()[0];
+            if (upperBound.equals(Object.class)) {
+                return true;
+            }
+        }
+        return false;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


benas:
Hello,

I've just fixed this exception here: https://github.com/benas/random-beans/commit/014783fd782817bc6db0f6e9e1c3a81b5722a3aa

I don't know why I have decided to throw an UnsupportedOperationException for raw types, I should have generated empty collections as discussed in #66 .. I think I'm tired.. I'll update this tomorrow :smile: 

Mahmoud

PascalSchumacher:
Hi Mahmound,

just saw you commit. :smile: 

With this request random-beans would also generates empty collections for things like `List<?>`, `Collection<?>`. There is no support for `Map`.

With the classpathScanning we could also fill collections with wildcards, but that requires more code to be written.

benas:
Hi,

Just fixed that, it will not wait for tomorrow! See here: https://github.com/benas/random-beans/commit/a81bb5644cbe1e643128bbc526e35012598c625b

Now, `List<?>`, `Collection<?>` and `Map<?>` are generated empty :wink: 

Cheers,
Mahmoud

random-beans/src/main/java/io/github/benas/randombeans/util/ReflectionUtils.java
```diff
@@ -165,6 +166,24 @@ public static boolean isParameterizedType(final Type type) {
     }
 
     /**
+     * Check if a type is an unbounded wildcard type
+     * @param parameterizedType the ParameterizedType to check
+     * @return true if the ParameterizedType is an unbounded wild card type, false otherwise
+     */
+    public static boolean isUnboundedWildcardType(final ParameterizedType parameterizedType) {
```
benas: Hey, this is better than the current static method `isWildcardType` of the ReflectionUtils class. But I think that testing if the type is instance of WildcardType is fair enough. Can you please try to use it in replacement of `isUnboundedWildcardType` and see if your tests pass? If so, it would be great to update the PR to keep only tests, they are more complete than the [poor](https://github.com/benas/random-beans/blob/master/random-beans/src/test/java/io/github/benas/randombeans/CollectionPopulationTest.java#L145) [ones](https://github.com/benas/random-beans/blob/master/random-beans/src/test/java/io/github/benas/randombeans/MapPopulationTest.java#L152) I've created .. Don't hesitate to get rid of them and their corresponding beans

PascalSchumacher: I think the `isWildcardType` method is good enough. It should identify all wildcards e.g. `<?>` `<? extends Runnable>`  The `isUnboundedWildcardType` checks if the wildcard is `<?>`. 

You solution is more generally useful, because it allows random-beans to generate empty collections for all collections with wildcard generics.

My solution is more limited, when I was writing it I was thinking to much about adding support for collections with bound wildcards to extends the classpath scanning.

I will reduce the pull request to the test this evening.

PascalSchumacher:
rebased test are now https://github.com/benas/random-beans/pull/73

PR #79, opened 2016-02-22
only load org.reflections classes when #scanClasspathForConcreteTypes…

PascalSchumacher:
… is enabled


benas:
Fantastic! Thank you for the quick update.

Do you think I can make the dependency optional after this change?

PascalSchumacher:
Yes you can make it optional

benas:
Ok great, I'll do it

PR #89, opened 2016-03-30
replace org.reflections with io.github.lukehutch.fast-classpath-scanner

PascalSchumacher:
fast-classpath-scanner has no dependencies and seems to be better maintained

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
         <joda-time.version>2.9.2</joda-time.version>
         <spring-context.version>4.2.4.RELEASE</spring-context.version>
         <objenesis.version>2.2</objenesis.version>
-        <reflections.version>0.9.10</reflections.version>
+        <fast-classpath-scanner.version>1.9.18</fast-classpath-scanner.version>
         <hibernate-validator.version>5.2.3.Final</hibernate-validator.version>
         <javax.el.version>2.2.6</javax.el.version>
         <mockito-core.version>1.10.19</mockito-core.version>
@@ -185,9 +185,9 @@
                 <version>${objenesis.version}</version>
             </dependency>
             <dependency>
-                <groupId>org.reflections</groupId>
-                <artifactId>reflections</artifactId>
-                <version>${reflections.version}</version>
+                <groupId>io.github.lukehutch</groupId>
+                <artifactId>fast-classpath-scanner</artifactId>
+                <version>${fast-classpath-scanner.version}</version>
                 <optional>true</optional>
             </dependency>
             <dependency>
--- a/random-beans/pom.xml
+++ b/random-beans/pom.xml
@@ -90,8 +90,8 @@
             <artifactId>objenesis</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.reflections</groupId>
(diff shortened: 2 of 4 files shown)


benas:
I was aware of Reflections library, but not fast-classpath-scanner. Looks interesting!

But with hindsight, do we really need a whole dependency for just 10 lines of code? Can we try to do it without this library and see if it really worth it?

I do believe there is a trade-off here and I try to make a pragmatic decision. If we really struggle with classpath scanning for concrete sub-types, then let's definitely go for using fast-classpath-scanner.

PascalSchumacher:
I haven't actually checked how much code is needed from Reflections/Fast-Classpath-Scanner. If it's really just a few lines I'm all for removing the dependency.

benas:
> If it's really just a few lines I'm all for removing the dependency.

Great! We are on the same page of the same philosophy

PascalSchumacher:
Related to this: I think it is also be possible to remove the commons-lang3 dependency, because random-beans uses just three of its methods.

PascalSchumacher:
I took a look at the implementation of fast-classpath-scanner.

We can not replace the dependency by just copying a few lines of code. Extracting the parts we need looks like a lot of work.

I do not think this is worth the effort for an optional dependency. 

Fast-classpath-scanner has no dependencies so it only adds a single 100k jar to the classpath of users that like to use random-beans with classpath scanning.

benas:
> We can not replace the dependency by just copying a few lines of code

This is not what I meant, sorry if I was not clear. The goal is not to extract the code as is from the library and remove the dependency. That would be a shame.

Looking at the changes, here is the only "feature" we are using from fast-classpath-scanner (correct me if I'm wrong):

```
 List<String> subTypes = type.isInterface() ? classpathScanner.getNamesOfClassesImplementing(type) : classpathScanner.getNamesOfSubclassesOf(type);
```

I thought about trying to write those 2 methods with vanilla java by ourselves (may be in a different way or even simpler than FCS) and not relying on a third party library. If we really struggle with it and found it takes so much effort, then adding the library makes sense.

That said, since you already took a look at the code of FCS and found it's not that easy, that's fine for me, I trust you. Now, after this little trade-off analysis, Let's make our life easier by using this library :smile:

> Fast-classpath-scanner has no dependencies so it only adds a single 100k jar

The size does not matter (I mean.. in this context :stuck_out_tongue_winking_eye:) if the library is worth it. and it does in our case. Thank you for suggesting FCS.

PascalSchumacher:
Sorry, it wasn't you that did not make himself clear but me. :(

What I meant by
>We can not replace the dependency by just copying a few lines of code

was: I think we can not copy or write a simplified version of the parts of fast-classpath-scanner that random-beans requires without significant effort (as far as I can tell, I'm not an expert at class-path-scanning :)).

Besides the two methods you mention we also require `classpathScanner.scan();`.

benas:
Ok that's fine don't worry.

I'm not expert neither, classpath scanning is tricky and it would be risky to deal with it for an optional dependency. Let's delegate that to FCS. I do believe we made a pragmatic decision after all :smile:

PR #91, opened 2016-03-31
remove commons-lang3 dependency in order to minimize the number of de…

PascalSchumacher:
…pendencies


benas:
Fantastic. less is more!

PR #92, opened 2016-03-31
merge random-beans-java8 module into random-beans module, now that we…

### LINKED ISSUE #88: Merge random-beans-java8 into random-beans?
Should random-beans-java8 be merged into random-beans now that random-beans requires Java 8?

PascalSchumacher:
… require at least java 8 (fixes #88)


benas:
Cool :+1:  I wish I have as much time as you to implement all these improvements!

I think we can still group all JSR 310 randomizer in a registry even though they reside in the core module. They form a logical group together and would best fit in their own registry. What do you think?

Alson the package name "java8" is fine, but would it be more appropriate to be more specific? May be "jsr310" or "threeten" (like [here](https://github.com/ThreeTen/threetenbp/tree/master/src/main/java/org/threeten))?

PascalSchumacher:
>I think we can still group all JSR 310 randomizers in a registry even though they reside in the core module. They form a logical group together and would best fit in their own registry. What do you think?

Makes sense. I will update the pull request.

>Also the package name "java8" is fine, but would it be more appropriate to be more specific? May be "jsr310" or "threeten" (like here)?

I agree. As it's only for java.time classes a more specific name is appropriate. I personally find "threeten" confusing. I would prefer "jsr310", or maybe it just "time" (similar to "collection")?

PascalSchumacher:
Reintroduced the separate registry and change the package name to "time". If you rather have it be "jsr310" or something else I will update the pull request.

benas:
Thanks for the update.

package "time" is fine. I think we need to move all (legacy) date and time randomizers inside it too.

I'll refactor randomizers to be grouped in logical packages. Here is a suggestion:

- collection: collection types randomizers (as is today)
- time: JSR 310 types randomizers (as in this PR) + move java.util.Date, java.util.Calendar, java.sql.Date, java.sql.Time and java.sql.Timestamp randomizers in this package
- number: numeric types randomizers
- text: textual randomizers
- misc: for others

PascalSchumacher:
Sounds good to me. :smile:

benas:
good. done.

PR #94, opened 2016-03-31
FakerBasedRandomizer: always use english locale if the no locale is p…

### LINKED ISSUE #86: random-beans-randomizers: Exception if java-faker does not support default locale
For example I get:

>com.github.javafaker.service.LocaleDoesNotExistException: de_DE could not be found, does not have a corresponding yaml file
	at com.github.javafaker.service.FakeValuesService.<init>(FakeValuesService.java:22)
	at com.github.javafaker.Faker.<init>(Faker.java:46)
	at io.github.benas.randombeans.randomizers.FakerBasedRandomizer.<init>(FakerBasedRandomizer.java:52)

I think it would be nice if random-beans would use the english locale in this case.

### LINKED ISSUE #93: random-beans-randomizers: different locale when using seed
When you create a FakerBasedRandomizer without a seed random-beans creates the faker object using the default constructor. This means that faker uses the english locale.

If you create a FakerBasedRandomizer with a seed random-beans passes the default locale into the faker constructor.

PascalSchumacher:
…assed into the constructor (fixes #86, fixes #93)

The other option is to always try to create the faker with the default locale and only use the english locale if a `LocaleNotFoundException` occurs.

### DIFF
--- a/random-beans-randomizers/src/main/java/io/github/benas/randombeans/randomizers/FakerBasedRandomizer.java
+++ b/random-beans-randomizers/src/main/java/io/github/benas/randombeans/randomizers/FakerBasedRandomizer.java
@@ -40,11 +40,11 @@ public abstract class FakerBasedRandomizer<T> extends AbstractRandomizer<T> {
     protected Faker faker;
 
     protected FakerBasedRandomizer() {
-        faker = new Faker();
+        faker = new Faker(Locale.ENGLISH);
     }
 
     protected FakerBasedRandomizer(final long seed) {
-        this(seed, Locale.getDefault());
+        this(seed, Locale.ENGLISH);
     }
 
     protected FakerBasedRandomizer(final long seed, final Locale locale) {


random-beans-randomizers/src/main/java/io/github/benas/randombeans/randomizers/FakerBasedRandomizer.java
```diff
@@ -40,11 +40,11 @@
     protected Faker faker;
 
     protected FakerBasedRandomizer() {
-        faker = new Faker();
+        faker = new Faker(Locale.ENGLISH);
```
PascalSchumacher: There is not difference between `faker = new Faker();` and `faker = new Faker(Locale.ENGLISH);`, but I think it's better to make it explicit, to show that the used locale is the same as when using a seed.

benas:
> (fixes #86, fixes #93)

Cool :smile:

two birds one stone

PR #95, opened 2016-03-31
create objects by using class.newInstance() and use objenesis only wh…

### LINKED ISSUE #84: "Pre-initialized" Bean Fields are nulled
When writing a dto, I often initialize collections to ensure they are never null. It appears that objenesis's approach to instantiation ends up producing an instance will null fields.Here's some code to show it. The first assert will pass, and the second will fail. Perhaps there is a perspective where this is desired behavior, but it was surprising to me, and not what I would prefer so I thought I would share.

`public class Main {
    public static void main(String[] args) {
        EnhancedRandomBuilder builder = new EnhancedRandomBuilder();
        builder.exclude(new FieldDefinition<>("myList", List.class, MyBean.class));

        assert new MyBean().getMyList().size() == 0;
        assert builder.build().nextObject(MyBean.class).getMyList().size() == 0;
    }

    public static class MyBean {
        private List<String> myList = new ArrayList<>();

        public List<String> getMyList() {
            return myList;
        }

        public void setMyList(List<String> myList) {
            this.myList = myList;
        }
    }
}
`

PascalSchumacher:
…en this fails (fixes #84)

I could not find a way to make it work with objenesis. 

I guess it's best to use the default constructor if it exists anyway.

### DIFF
--- a/random-beans/src/main/java/io/github/benas/randombeans/ObjectFactory.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/ObjectFactory.java
@@ -50,18 +50,24 @@ class ObjectFactory {
     private boolean scanClasspathForConcreteTypes;
 
     <T> T createInstance(final Class<T> type) {
-        T result;
         if (scanClasspathForConcreteTypes && isAbstract(type)) {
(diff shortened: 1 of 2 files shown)


benas:
Great. Thank you for this fix.

Issue #28, opened 2015-07-31
Allow to use Supplier from java8 and guava as Randomizer


Toilal:
Create a new PopulatorBuilder method supporting Supplier instead of Randomizer.

```java
class PopulatorBuilder8 extends PopulatorBuilder {
    public PopulatorBuilder registerRandomizer(final Class type, final Class fieldType, final String fieldName, final Supplier<?> randomizer) {
        return super.registerRandomizer(type, fieldType, fieldName, Java8Adapters.adapt(randomizer)):
    }
}
```

Same idea for guava Supplier.

It requires #26 to keep support of java version. (jpopulator-java8 and jpopulator-guava modules)

benas:
Indeed, great idea :+1:  Thank you Rémi!

Let's plan that for next release :smiley:

benas:
I was trying to add support for `Supplier` in the Builder but this would make the library require Java 8 as minimum Java version (while some people are still using Java 6, example in #43 ) .

I was not able to use java source level to 1.8 and target level to 1.6 (because only `random-beans-java8` module requires Java 8). So I need to pay attention to not use Java 8 APIs outside this module..

As of today, the library is compiled with Java 8 but the binary is still Java 6 compatible.

PascalSchumacher:
I have started work an a pull request to add support for java 8 suppliers. It should be complete in a few days.

PascalSchumacher:
pull request which adds the feature to use a java8 supplier: https://github.com/benas/random-beans/pull/96

benas:
Hi @Toilal 

I've added support to use Java 8 suppliers as randomizers.

It would be great to give it a try if possible to see if it corresponds to what you have requested.
Many thanks upfront.

Kind regards
Mahmoud

Toilal:
I think it's good thanks ! I'll effectively try later.

benas:
Hi Rémi,

No problem, take your time.

This will be released soon in v3.0.0
In case of any issue please file a bug.

Note that support for the standard Java 8 API `Supplier` has been added. Support for guava API will not be included since it is specific.

Kind regards
Mahmoud

PR #96, opened 2016-03-31
Allow registration of a Supplier of Randomizers #28

PascalSchumacher:
Just a quick prototype and not exactly what was requested in #28, but maybe good enough?

### DIFF
--- a/random-beans/src/main/java/io/github/benas/randombeans/EnhancedRandomBuilder.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/EnhancedRandomBuilder.java
@@ -31,6 +31,7 @@ import io.github.benas.randombeans.randomizers.misc.SkipRandomizer;
 import io.github.benas.randombeans.randomizers.registry.CustomRandomizerRegistry;
 
 import java.util.*;
+import java.util.function.Supplier;
 
 import static io.github.benas.randombeans.util.Constants.DEFAULT_SEED;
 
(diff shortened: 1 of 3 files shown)


PascalSchumacher:
updated

benas:
Hello Pascal,

Thanks for implementing this feature. I think the idea of #28 is to use Suppliers as Randomizers, ie, Suppliers will supply random values directly and not supply randomizers.

Looking at this change:

```
public <T, F, R> EnhancedRandomBuilder randomize(FieldDefinition<T, F> fieldDefinition, Supplier<Randomizer<R>> randomizerSupplier)
```
Instead, I guess this would be 

```
public <T, F, R> EnhancedRandomBuilder randomize(FieldDefinition<T, F> fieldDefinition, Supplier<R> randomizerSupplier)
```

This is quite similar to the API suggested in #28. Do you see the difference?

> Just a quick prototype and not exactly what was requested in #28, but maybe good enough?

I saw you changes and I believe this is good enough. But I've got another idea on how to implement this and I would like to know what do you think about it. 

Instead of having two mappings (one for randomizers and one for suppliers):

```
private Map<FieldDefinition<?, ?>, Randomizer<?>> customRandomizersRegistry = new HashMap<>();
private Map<FieldDefinition<?, ?>, Supplier<?>> customRandomizersSupplierRegistry = new HashMap<>();
```

and the corresponding logic to both of them, my idea is to dynamically create a proxy randomizer for the supplier provided by the user. This way, everything remains intact, we only need to adapt the Supplier into a Randomizer. Here is a quick prototype:

```java
import io.github.benas.randombeans.api.Randomizer;

import java.lang.reflect.InvocationHandler;
import java.lang.reflect.Method;
import java.lang.reflect.Proxy;
import java.util.function.Supplier;

class RandomizerProxy implements InvocationHandler {

    private Object target;

    public RandomizerProxy(final Object target) {
        this.target = target;
    }

    public static Randomizer asRandomizer(final Supplier supplier) {
        return (Randomizer) Proxy.newProxyInstance(
                Randomizer.class.getClassLoader(),
                new Class[]{Randomizer.class, Comparable.class},
                new RandomizerProxy(supplier));
    }

    @Override
    public Object invoke(final Object proxy, final Method method, final Object[] args) throws Throwable {
        Method get = target.getClass().getMethod("get");
        if ("getRandomValue".equals(method.getName())) {
            return get.invoke(target, args);
        }
        return null;
    }
}
```

and a test:

```java
import io.github.benas.randombeans.api.Randomizer;
import org.junit.Test;

import java.util.function.Supplier;

import static org.assertj.core.api.Assertions.assertThat;

public class RandomizerProxyTest {

    @Test
    public void testAsRandomizer() throws Exception {
        MySupplier supplier = new MySupplier();
        Randomizer randomizer = RandomizerProxy.asRandomizer(supplier);
        Object randomValue = randomizer.getRandomValue();

        assertThat(randomValue).isInstanceOf(String.class).isEqualTo("foo");
    }

    class MySupplier implements Supplier<String> {

        @Override
        public String get() {
            return "foo";
        }
    }
}
```

At registration time, we need to call the `RandomizerProxy.asRandomizer(supplier)` method with the supplier provided by the user. Do you see the idea?

This way, we can register Suppliers and Randomizers in a consistent way. Random Beans will create a dynamic proxy for the Supplier at runtime. This technique will allow us to add support for guava supplier or any other type easily without impacting the internal logic.

I would love to hear your thoughts.
Kind regards
Mahmoud

PascalSchumacher:
Hi Mahmoud,

like your idea a lot. :smile: 

It's exactly what was requested in the issue and simple to use.

I was struggling to come up with a way to use a supplier as a randomizer, but you solved it elegantly. :smile:

Kind regards,
Pascal

Issue #82, opened 2016-02-29
Still can not easily register Class randomizer programatically


wheredevel:
Hi,

I'm still struggling with registering Randomizer for a certain Type programatically.
Wiki describes randomizer registration for: field... named..of class.
While I need: just class.

Please, help. Did I miss the right line in a manual?

benas:
Hi,

The short answer is : You can provide a [custom registry](https://github.com/benas/random-beans/wiki/Grouping-Randomizers) in which you associate a randomizer to a type.

Now this is a design question: As a user of the library, I want to be able to register a custom randomizer if the field:

* has a given type
* has a given name
* is declared in a given class
* is annotated with a given annotation
* (two, three or) All of the above

From an API design point of view, there should be a way to define how to identify a field in the process of bean introspection and this is the role of the `FieldDefinition` class.

Field name, type and declaring class is the common denominator to identify a field. You can not have two fields with the same name, same type and same declaring class. Field definition are [unique](https://github.com/benas/random-beans/blob/master/random-beans/src/main/java/io/github/benas/randombeans/FieldDefinition.java#L68) according to these three criteria. **For the moment**, `FieldDefinition` and its builder are designed to cover these criteria.

You can still use registries to provide custom randomizers by `Field` or by `Type`. When you implement a custom `RandomizerRegistry`, you get access to the `Field` being introspected and here you can get all informations above: field name, type, annotations etc and hence provide your custom field -> randomizer mapping.

Hope this helps
Kind regards
Mahmoud

wheredevel:
Everything sounds right.
I'd only emphasize the existence of 2 API use-cases:
1. Building a specific randomizer - while specifying what registries to use, etc.
2. Invoking (a previously built) randomizer on a Class<?> - while modifying the randomization behavior during this particular randomization invocation only.

Case 1 is well covered.

Case 2 - there we can modify using FieldDefinition, which covers most cases.
Still, I had this edge case, when ALL the fields of a particular Type would have to be generated differently from Randomizer default. And FieldDefinition couldn't do the job for me. I wonder if there're should be a way - or is it just too much (for example, because the modification logic becomes too complex.)

PascalSchumacher:
Possible fix: https://github.com/benas/random-beans/pull/97

benas:
Hi @wheredevel 

@PascalSchumacher has kindly added support for registering a randomizer for a given type. This removes the need to implement a custom registry for that.

[...]

PR #97, opened 2016-04-01
allow registration of a randomizer for a type without implementing a …

PascalSchumacher:
…registry #82 

Just a quick prototype to show how #82 could be done.

@benas No problem if you think that this should not be added to random-beans. :smile:


benas:
That's great! Thanks again.

Issue #29, opened 2015-07-31
Mockito like API


Toilal:
It would be awesome to have some kind of [Mockito](http://mockito.org/)-like API to register randomizer using reflection. It involves some hacking with [cglib](https://github.com/cglib/cglib), but it's possible to implement this kind of magic ...

```java

import static io.github.benas.jpopulator.api.reflect.*

public static class BeanA {
    private String label;
    private int id;

    // getters/setters are defined ...
}

...
BeanA a = jpopulator(BeanA.class)

randomizer(a.getLabel(), new Randomizer() { 
    // The randomizer ...
});
// instead of PopulatorBuilder.registerRandomizer(BeanA.class, String.class, "label", ...);

ignore(a.getId()); // Ignore this field

BeanA populatedBean = populate(a); // Generate a single bean
List<BeanA> populatedList = populate(a, 100); // Generate a list of 100 beans

```

This API would make jPopulator based tests really easier to maintain on bean changes, because there are no hardcoded classes and property names in randomizer registrations.

benas:
This is a nice to have feature. A Mockito-like API has been also suggested [here](https://github.com/benas/jPopulator/issues/9#issuecomment-62683160) but as you said, it requires some black magic and a big refactoring of the code base. Let's keep this feature request in the pipe. Contributions are welcome!

Toilal:
I'm not sure it requires refactoring, I think this "black magic" API could delegate to the existing API.

There are some explanations on how Mockito `when()` on [stackoverflow](http://stackoverflow.com/questions/14440270/how-does-mockito-when-invocation-work). For me, It's still quite confuse on how to implement this cleanly though, but i'll give it a try next week.

PascalSchumacher:
For a quick and very unfinished prototype see: https://github.com/benas/random-beans/pull/98

benas:
Hi All,

I need some time to analyze in depth the relevance and effort needed for this PR.
Hence, I'm postponing it to v3.1

Kind regards
Mahmoud

punkratz312:
+1

benas:
Even though the requested feature is a good idea, I think the added value of the syntactic sugar is not worth it in comparison with the effort needed to implement it. I'm closing this issue for now.

PR #98, opened 2016-04-02
prototype for mokito like api #29

PascalSchumacher:
I hacked together a quick and very unfinished prototype.

Naming is obviously completely wrong.

### DIFF
--- a/random-beans/pom.xml
+++ b/random-beans/pom.xml
@@ -92,6 +92,11 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+           <groupId>net.bytebuddy</groupId>
+           <artifactId>byte-buddy</artifactId>
+           <version>1.4.4</version>
+        </dependency>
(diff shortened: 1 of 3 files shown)


benas:
Thank you for this PR. Happy to see byte buddy here :-)

Please give me some time to take a look at your changes and how this technique works behind the scene.

PascalSchumacher:
Sure. No need to hurry. :)

Also please feel free to reject. I'm not sure this feature is worth the effort it will require.

PascalSchumacher:
The more I think about it the less I'm convinced that this feature is worth the considerable implementation effort it requires.

benas:
I do agree. I think the added value of the syntactic sugar is not worth it in comparison with the effort needed to implement it. Although it is a good idea, I think we can discard it for now. I'm closing this PR along with the original feature request ( #29 )

PR #102, opened 2016-04-03
More randomizers

PascalSchumacher:


PascalSchumacher:
Failed tests:  BeanValidationTest.generatedBeanShouldBeValidUsingBeanValidationAPI:94 

Expecting empty but was:<[ConstraintViolationImpl{interpolatedMessage='must be in the past', propertyPath=birthday, rootBeanClass=class io.github.benas.randombeans.validation.BeanValidationAnnotatedBean, messageTemplate='{javax.validation.constraints.Past.message}'}]>

  BeanValidationTest.generatedValuesShouldBeValidAccordingToValidationConstraints:64 

Expecting:
 <2025-10-05T08:40:19.401>
to be in the past but was not.


Seems unrelated to my changes, as far as I can tell.

PascalSchumacher:
Test failure was caused by my commit which move the pre-java-8 time related randomizers to `TimeRandomizerRegistry`. I remove that commit from the pull request.

random-beans/src/main/java/io/github/benas/randombeans/randomizers/misc/EnumRandomizer.java
```diff
@@ -32,7 +32,7 @@
  *
  * @author Mahmoud Ben Hassine (mahmoud.benhassine@icloud.com)
  */
-public class EnumRandomizer<T extends Enum<T>> extends AbstractRandomizer<Enum<T>> {
+public class EnumRandomizer<T> extends AbstractRandomizer<T> {
```
benas: Hi,

I'm not sure I understand the purpose of removing this bound. This will allow to have `EnumRandomizer<Integer> ir = new EnumRandomizer<>(Integer.class);` which makes no sense. The idea was to restrict generic types to only enums for `EnumRandomizer`. You can still write `EnumRandomizer<Month> mr = new EnumRandomizer<>(Month.class);` even with the bounded generic type since `Month` is an enum. Do you agree?

Cheers
Mahmoud

PascalSchumacher: Hi,

I completely agree.

I changed the old generic signature because it was "wrong"/ugly to use. The new one was less verbose for the user (see other changes of the same commit), but as you said it was too permissive.

I have update the signature again. The new version combines the best of both versions. It's nice to use, but requires the parameter to be a `Enum` with the right type.

Thanks for the review. :)

Cheers,
Pascal

benas:
> I changed the old generic signature because it was "wrong"/ugly to use. The new one was less verbose for the user (see other changes of the same commit), but as you said it was too permissive.

Yes. Clearly, `Randomizer<Gender> randomizer` is better than `Randomizer<Enum<Gender>> randomizer;`.

I also like the idea of replacing `MonthRandomizer` by `EnumRandomizer<Month>`. :+1: 

Thank you for these nice additions.

PR #103, opened 2016-04-06
Fix Bug and Improve test coverage

PascalSchumacher:


benas:
Hi,

Well spotted for the EnumRandomizer seed bug :+1: 

In regards to improving test coverage, I think we can keep the current test structure and call the static factory method instead of the default constructor. This way, we improve test coverage without changing the structure. I like to see the declaration of the tested object, it's setup and the test separately:

```java
 public class SkipRandomizerTest {
  
     private SkipRandomizer skipRandomizer;
 
     @Before
     public void setUp() {
         skipRandomizer = aNewSkipRandomizer();
     }
 
      @Test
      public void generatedValueShouldBeNull() {
         assertThat(skipRandomizer.getRandomValue()).isNull();
      }
 
  }
```

Even better, I was planning to refactor all tests to follow this pattern when the test class contains several different setups and tests (like [here](https://github.com/benas/random-beans/blob/master/random-beans/src/test/java/io/github/benas/randombeans/randomizers/range/BigDecimalRangeRandomizerTest.java)):

```java
 public class BigDecimalRangeRandomizerTest {
  
     private BigDecimalRangeRandomizer randomizer;
 
      @Test
    public void generatedValueShouldBeWithinSpecifiedRange() {
        //Given
        randomizer = new BigDecimalRangeRandomizer(1L, 10L);

        //When
        BigDecimal randomValue = randomizer.getRandomValue();

        //Then
        assertThat(randomValue.longValue()).isBetween(1L, 10L);
    }

    @Test
    public void whenSpecifiedMinValueIsNullThenShouldUseDefaultMinValue() {

        //Given
        randomizer = new BigDecimalRangeRandomizer(null, 10L);

        //When
        BigDecimal randomValue = randomizer.getRandomValue();

        //Then
        assertThat(randomBigDecimal.longValue()).isLessThanOrEqualTo(max);
        
    }
 
  }
```

There are already some tests that follow this pattern and the idea is to make all of them coherent. What do you think?

benas:
Changing tests to follow the Given/When/then pattern is not the goal this PR. I'll do it later.

If you can just keep the current tests structure, it would be fine to merge.

Many thanks upfront

PascalSchumacher:
In my opinion

```java
public class SkipRandomizerTest {

     private SkipRandomizer skipRandomizer;

     @Before
     public void setUp() {
         skipRandomizer = aNewSkipRandomizer();
     }

      @Test
      public void generatedValueShouldBeNull() {
         assertThat(skipRandomizer.getRandomValue()).isNull();
      }
  }
```

is to verbose. It's too much effort to write and read for such a simple test.

I only use a `setUp` method if several methods reuse the same set-up or if there is a lot to set-up and it is not relevant to the test.

Changing the tests to the current structure does not make sense imho, because they share no set-up.

I can change each test to a thee line structure e.g.

```java
 @Test
 public void generatedListShouldNotBeEmpty() {
     ListRandomizer listRandomizer = aNewListRandomizer(elementRandomizer);

     List<String> strings = listRandomizer.getRandomValue() 
    
     assertThat(strings).isNotEmpty();
}
```

if you prefer that.

benas:
In my current work we are using [unitils](http://www.unitils.org/tutorial-inject.html). I like the `@TestedObject` annotation to make it clear which object is under test, especially when you have many declarations with mocks, stubs, etc. That's why I said that I like to see separate declaration, setup and test. But this should not be systematic, I do agree with you! In random beans, I'm not planning to use unitils. The test class name should already give an idea about the object being tested.

Anyway, your three parts suggestion is also fine. It follows the Given/When/Then pattern. There are already [some tests](https://github.com/benas/random-beans/blob/master/random-beans/src/test/java/io/github/benas/randombeans/EnhancedRandomImplTest.java#L265) that follow this pattern, and I'm planning to refactor all for them to be coherent. May be adding some comments would make the test clearer:

```java
 @Test
 public void generatedListShouldNotBeEmpty() {
     // Given
     ListRandomizer listRandomizer = aNewListRandomizer(elementRandomizer);

     // When
     List<String> strings = listRandomizer.getRandomValue() 

     // Then
     assertThat(strings).isNotEmpty();
}
```

Go for it. Thank you Pascal

BTW: I've seen some folks using [labels](http://insightfullogic.com/2015/Jan/19/given-when-then-java/) for that :wink:

PascalSchumacher:
I think:

```java
 @Test
 public void generatedListShouldNotBeEmpty() {
     // Given
     ListRandomizer listRandomizer = aNewListRandomizer(elementRandomizer);

     // When
     List<String> strings = listRandomizer.getRandomValue() 

     // Then
     assertThat(strings).isNotEmpty();
}
```
does not add much over:

```java
 @Test
 public void generatedListShouldNotBeEmpty() {
     ListRandomizer listRandomizer = aNewListRandomizer(elementRandomizer);

     List<String> strings = listRandomizer.getRandomValue() 

     then(strings).isNotEmpty();
}
```

but that's probably just my laziness ;). I will use use the schema you suggested.

benas:
There is no difference, except the three explicit comments as I said:

>  May be adding some comments would make the test clearer:

But don't bother yourself with this, I'll add those comments later in my planned refactoring.

PascalSchumacher:
>But don't bother yourself with this, I'll add those comments later in my planned refactoring.

No problem. I have already started to do so. :smile: 

As a fun experiment I merged ListRandomizer, QueueRandomizer and SetRandomizer into a single parameterized class, see: https://github.com/benas/random-beans/pull/103/commits/00c189fb395a5cc37922457730924350dc3e2c87

PascalSchumacher:
>As a fun experiment I merged ListRandomizer, QueueRandomizer and SetRandomizer into a single parameterized class, see: 00c189f

These tests are probably not best example and the resulting test is a bit complex, but https://github.com/benas/random-beans/pull/103/commits/122a62a36064852b2ac1f9f0f828d463d5c8ba7a is a good example of the power of parametrized tests.

random-beans/src/test/java/io/github/benas/randombeans/randomizers/number/AtomicLongRandomizerTest.java
```diff
@@ -40,12 +39,10 @@ public void setUp() throws Exception {
...
     }
 
-    @Ignore("AtomicLong are not intended to be comparable, see http://stackoverflow.com/questions/7567502/why-are-two-atomicintegers-never-equal")
     @Test
     public void shouldGenerateTheSameValueForTheSameSeed() {
         AtomicLong actual = randomizer.getRandomValue();
-        AtomicLong expected = new AtomicLong(-5106534569952410475L);
 
-        assertThat(actual).isEqualTo(expected);
+        assertThat(actual.get()).isEqualTo(-5106534569952410475L);
```
benas: Thank you for fixing this!!

benas:
Great! It's very powerful indeed: reducing a dozen of tests into a single parametrized one is clearly elegant. A couple of questions since I've never used junit-dataprovide before:

* Does junit-dataprovide requires those toString method you have added?
* Is there a naming convention between methods that provide data and methods that test the code? For instance: `@DataProvider public static Object[] generatedNumberShouldNotBeNull()` and `@Test @UseDataProvider public void generatedNumberShouldNotBeNull(Randomizer<?> randomizer)` should have the same name?

In your experience, is it better than built-in support for [parametrized tests](https://github.com/junit-team/junit4/wiki/Parameterized-tests) in JUnit? May be in our case the built-in feature is enough?

PascalSchumacher:
> Does junit-dataprovide requires those toString method you have added?

No it does not require a `toString`, but `toString` is used to display the parameter values in the test names.

> Is there a naming convention between methods that provide data and methods that test the code? For instance: @DataProvider public static Object[] generatedNumberShouldNotBeNull() and @Test @UseDataProvider public void generatedNumberShouldNotBeNull(Randomizer<?> randomizer) should have the same name?

That a supported naming convention. It's also possible to use an explicit reference e.g.: `@UseDataProvider("dataProviderMethodName")`.

>In your experience, is it better than built-in support for parametrized tests in JUnit? May be in our case the built-in feature is enough?

One advantage over the build-in support is that you can have multiple parametrized tests in a single class.

There are other nice features e.g.: https://github.com/TNG/junit-dataprovider/wiki/Features#string-syntax

benas:
ok, it makes sense.

Thank you for this update. Great, as usual :smile:

PascalSchumacher:
>Great, as usual :smile:

Thanks and thanks for merging. :smile:

PR #111, opened 2016-04-13
update to java-faker 0.8

PascalSchumacher:
the 0.8 release supports many more locales, but it also uses different data-files, so some tests had to be changed


benas:
Thx for the update

Issue #110, opened 2016-04-13
FullNameRandomzier vs NameRandomizer


PascalSchumacher:
Both generate the same data, because in java-faker `fullName()` is just an an alias for `name()`

```java
   public String fullName() {
        return name();
    }
```
I think we should remove one of the randomizers.

What do you think?

benas:
yes we can keep FullNameRandomizer and remove NameRandomizer.
This way we have: FirstName, LastName and FullName randomizers.

PR #114, opened 2016-04-14
delete NameRandomizer which is a duplication of FullNameRandomzier (c…

PascalSchumacher:
…loses #110)


benas:
Hey, thanks. No more duplicates :smile:

Issue #113, opened 2016-04-14
First Collection without seed empty


punkratz312:
My Class has 4 fields:

        String string;

        Collection<String> strings;

        Collection<String> strings2;

        Collection<String> strings3;

without seed the first collection "strings" is empty. 

I´ve debugged the code a little. It was quite confusing that ArrayPopulator.getRandomPrimitiveArray (59) randomSize is 0 if its the first collection (enhancedRandom.nextInt() returns 0). Hopefully thats a useful hint ^^.

![screen shot 2016-04-14 at 15 11 46](https://cloud.githubusercontent.com/assets/8830888/14529673/f6ed7fa2-0255-11e6-8158-2e39ee13d7df.png)
![screen shot 2016-04-14 at 15 24 08](https://cloud.githubusercontent.com/assets/8830888/14529674/f6f21404-0255-11e6-9bc7-3e897c3a09f0.png)

punkratz312:
But that not even my original problem, I have issues with the crazy chines chars. I´am serializing the Bean and get some issues. Why aren´t the strings of the collection similar to the simple field String?

PascalSchumacher:
>I´ve debugged the code a little. It was quite confusing that ArrayPopulator.getRandomPrimitiveArray (59) randomSize is 0 if its the first collection (enhancedRandom.nextInt() returns 0). 

I just took a look at the code. If you do not pass a seed random-beans always uses the same default seed: 1234567890L. For this reason the generated values are not really random. In each execution the same sequence of numbers will be generated. I think it would be better if random-beans was changed to use a random seed.

>I have issues with the crazy chines chars.

Looks like the strings also contain non-printable/control characters. I guess random-beans should stick to printable characters (maybe it's a good idea to make this configurable?).

punkratz312:
Nice approach, fixed the problem: (First 2.0.0, Second 3.0.0-SNAPSHOT)
![screen shot 2016-04-15 at 11 15 01](https://cloud.githubusercontent.com/assets/8830888/14557121/f178a276-02fb-11e6-9ff3-58a951915563.png)
![screen shot 2016-04-15 at 11 18 50](https://cloud.githubusercontent.com/assets/8830888/14557122/f18f322a-02fb-11e6-9ec2-1805c187c1e8.png)

punkratz312:
Would it be possible to change, or at lease make it configureable to change these chinese characters.

[...]

PR #115, opened 2016-04-14
replace constant default seed by random seed to make the generated va…

PascalSchumacher:
…lues random again (#113)

### DIFF
--- a/random-beans/src/main/java/io/github/benas/randombeans/EnhancedRandomBuilder.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/EnhancedRandomBuilder.java
@@ -34,7 +34,6 @@ import java.util.*;
 import java.util.function.Supplier;
 
 import static io.github.benas.randombeans.RandomizerProxy.asRandomizer;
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
If this is merged I think we should expose a method for the user to determine the seed used by `EnhancedRandom`.

benas:
Good idea indeed, that's more random..

punkratz312:
Nice approach, fixed the problem: (First 2.0.0, Second 3.0.0-SNAPSHOT)
screen shot 2016-04-15 at 11 15 01
screen shot 2016-04-15 at 11 18 50

benas:
> Good idea indeed, that's more random..

The build is also random :smile: It's failing now ..

punkratz312:
WorX on my machine (dad34b18b70a34ab54533fd9bf76f653a26e65a9) mvn clean install 8 times always success

benas:
likewise here. But I succeeded to make it fail after 20 attempts..

the build was green on the PR: https://travis-ci.org/benas/random-beans/builds/123129961
but fails after the merge: https://travis-ci.org/benas/random-beans/builds/123257530

a bit weird, I'll check that

PR #119, opened 2016-04-20
quick and dirty prototype to allow users to define charset for strings

PascalSchumacher:


benas:
I always begin with the simplest solution that could possibly work :wink: 

Your changes are fine, but they impact only arrays. String and Character fields will not use the specified charset. What if we add the charset as parameter in the constructor of `StringRandomizer` and `CharacterRandomizer` ?

PascalSchumacher:
Sure, but I'm not actually sure this is the best way to do it. Let's release 3.0 first and maybe add this feature later on, what do you think?

benas:
Yes, the problem is that the `ArrayPopulator` does not reuse existing randomizers. It should delegate value generation to registered randomizers by type. This way, if the user registers a CharacterRandomizer with a charset (and probably a seed) at construction time, both char arrays and char fields use the same randomizer. do you see the point?

I think if we:
* make the `ArrayPopulator` delegate value generation to existing randomizers
* and add Charset param to CharacterRandomizer and StringRandomizer

it would be fine for 3.0. This does not change the builder API, just a refactoring and new param in constructors.

benas:
Hi,

I've made the aforementioned refactoring. Now, the array populator uses registered randomizers (it gets them through the RandomizerProvider).

the only remaining thing is to replace my quick and dirty solution (with [alphabet](https://github.com/benas/random-beans/blob/master/random-beans/src/main/java/io/github/benas/randombeans/randomizers/text/CharacterRandomizer.java#L75)) with your elegant solution using a charset. The `StringRandomizer` should also be updated.

The user will be able to register a `CharacterRandomizer` with a charset at construction time and char arrays and fields will be randomized with characters from the specified charset.

Cheers
Mahmoud

PascalSchumacher:
Hi Benas,

thanks!

I reduced the pull request to a CharacterUtils class.

PascalSchumacher:
I can also do the integration, but that will have to wait a few days.

So if you like to do it, feel free.

Cheers,
Pascal

benas:
Hi,

Thanks for the update! You've already done 80% of the work 😄.
No worries, I'll do the integration.

Cheers
Mahmoud

Issue #117, opened 2016-04-18
rename nextObjects method into objects to conform with the standard Random API naming pattern


benas:
Starting from Java 8, the `Random` class added methods that return a stream of int, double, etc (called `ints`, `doubles`) in addition to the existing `nextInt`, `nextDouble`,.. methods.

The `EnhancedRandom` provided by Random Beans should follow this pattern and provide a method (called `objects`) that return a stream of objects in addition to the `nextObject` method.

Here is the new signature (replacement of `List<T> nextObjects` method):

```java
public abstract class EnhancedRandom extends Random {

    /**
     * Generate a stream of random instances of the given type.
     *
     * @param type           the type for which instances will be generated
     * @param excludedFields the name of fields to exclude
     * @param streamSize     the number of instances to generate
     * @param <T>            the actual type of the target objects
     * @return a stream of random instances of the given type
     * @throws ObjectGenerationException when unable to populate an instance of the given type
     */
    public abstract <T> Stream<T> objects(final Class<T> type, final int streamSize, final String... excludedFields);

}
```

PR #122, opened 2016-04-25
add static factory method to generate a random beanS with default con…

punkratz312:
…figuration parameters

### DIFF
--- a/random-beans/src/main/java/io/github/benas/randombeans/api/EnhancedRandom.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/api/EnhancedRandom.java
@@ -73,13 +73,28 @@ public abstract class EnhancedRandom extends Random {
     /**
      * Generate a random instance of the given type.
      *
-     * @param type the target class type
+     * @param type           the target class type
      * @param excludedFields the name of fields to exclude
-     * @param <T>  the target type
+     * @param <T>            the target type
      * @return a random instance of the given type
      */
     public static <T> T random(final Class<T> type, final String... excludedFields) {
         return aNewEnhancedRandomBuilder().build().nextObject(type, excludedFields);
+    }
+
+    /**
+     * Provides static access to {@link EnhancedRandom#objects(Class, String...)}
+     */
+    public static <T> Stream<T> randoms(final Class<T> type, final String... excludedFields) {
+        EnhancedRandom enhancedRandom = aNewEnhancedRandomBuilder().build();
+        return enhancedRandom.objects(type, enhancedRandom.nextInt(100), excludedFields);
+    }
+
+    /**
+     * Provides static access to {@link EnhancedRandom#objects(Class, int, String...)}
+     */
+    public static <T> Stream<T> randoms(final Class<T> type, final int streamSize, final String... excludedFields) {
+        return aNewEnhancedRandomBuilder().build().objects(type, streamSize, excludedFields);
     }
 
 }
--- a/random-beans/src/main/java/io/github/benas/randombeans/api/EnhancedRandom.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/api/EnhancedRandom.java
@@ -84,6 +84,8 @@ public abstract class EnhancedRandom extends Random {
 
     /**
      * Provides static access to {@link EnhancedRandom#objects(Class, String...)}
+     *
+     * FIXME randoms is not a good name - maybe extract it into StreamEnhancedRandom or something
      */
     public static <T> Stream<T> randoms(final Class<T> type, final String... excludedFields) {
         return aNewEnhancedRandomBuilder().build().objects(type, excludedFields);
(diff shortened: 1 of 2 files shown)


punkratz312:
I´m not very happy about the name "randoms". Hopefully we could find a better one.

random-beans/src/main/java/io/github/benas/randombeans/api/EnhancedRandom.java
```diff
@@ -73,13 +73,28 @@
...
+     */
+    public static <T> Stream<T> randoms(final Class<T> type, final String... excludedFields) {
+        EnhancedRandom enhancedRandom = aNewEnhancedRandomBuilder().build();
+        return enhancedRandom.objects(type, enhancedRandom.nextInt(100), excludedFields);
+    }
+
+    /**
+     * Provides static access to {@link EnhancedRandom#objects(Class, int, String...)}
+     */
+    public static <T> Stream<T> randoms(final Class<T> type, final int streamSize, final String... excludedFields) {
```
benas: great! since you're not happy with the name "randoms", we can simply keep it "random" to overload the single instance generation method?

random-beans/src/main/java/io/github/benas/randombeans/api/EnhancedRandom.java
```diff
@@ -73,13 +73,28 @@
...
-     * @param <T>  the target type
+     * @param <T>            the target type
      * @return a random instance of the given type
      */
     public static <T> T random(final Class<T> type, final String... excludedFields) {
         return aNewEnhancedRandomBuilder().build().nextObject(type, excludedFields);
     }
 
+    /**
+     * Provides static access to {@link EnhancedRandom#objects(Class, String...)}
```
benas: This is not exactly the same, since the `EnhancedRandom#objects(Class, String...)` returns an infinite stream and not a bounded one (with 100 elements). I think this method can be removed and keep the user choose the size of the stream with `randoms(final Class<T> type, final int streamSize, final String... excludedFields)`. what do you think?

benas:
Don't hesitate to add yourself to the contributors list : https://github.com/benas/random-beans/blob/master/README.md#awesome-contributors

thank you for this PR!

random-beans/src/main/java/io/github/benas/randombeans/api/EnhancedRandom.java
punkratz312: This does not work. Do I get you wrong here?

screen shot 2016-04-25 at 18 19 35

punkratz312: The parameter signature is the same, I´m pertty sure the compiler can not determine which method should be used. (The returnType is not enough)

random-beans/src/main/java/io/github/benas/randombeans/api/EnhancedRandom.java
punkratz312: infinite stream - wow. that explains why the test does not return ^^

punkratz312: I can´t imagine an usecase, but im sure there is one ^^, of an infinite stream.

benas:
> infinite stream - wow. that explains why the test does not return ^^

That's why I've [limited it to 10](https://github.com/benas/random-beans/blob/master/random-beans/src/test/java/io/github/benas/randombeans/EnhancedRandomImplTest.java#L134) in the test 😉 

> I can´t imagine an usecase, but im sure there is one ^^, of an infinite stream.

That was added to be consistent with standard Random API that random beans extends. More details [here](https://github.com/benas/random-beans/issues/117)

In regards to the overloading `random` method that I've suggested, I'm sorry, it won't work. So we can just keep the last one: 

```
public static <T> Stream<T> random(final Class<T> type, final int streamSize, final String... excludedFields) { ... }
```

Could you please update the PR? It will be fine to merge

punkratz312:
I think its more consistent if we add the s (plurar) to random for both stream methods.

Maybe we can extract the Stream returning methods.

benas:
Thank you for this update!

with hindsight, you are right. I don't see a real added value of a method returning an infinite stream of random objects. the user can always specify the stream size. I have added it to be consistent with [existing](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html#ints--) [methods](https://docs.oracle.com/javase/8/docs/api/java/util/Random.html#longs--) in the Random API.

That said, I'll remove it. That's confusing :smile:

punkratz312:
I´m not sure about the workflow. Is the SNAPSHOT (including this change) automatically deployed to mvn central?

I could not use the new API, our project depends on 3.0.0-SNAPSHOT.

Thanks.

benas:
> Is the SNAPSHOT (including this change) automatically deployed to mvn central?

No, I do it manually. I've just re-uploaded the latest snapshot version. It should be fine now.

punkratz312:
Ah ok, thanks a lot :). You make me happy as always.

PR #123, opened 2016-04-25
Replace boilerplate with lombok

punkratz312:


punkratz312:
Nice feature!

benas:
Coool! Less is more :smile: 

Can you please resync with the master branch?

random-beans/pom.xml
```diff
@@ -101,6 +101,12 @@
             <artifactId>junit-dataprovider</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.16.8</version>
+            <scope>provided</scope>
```
benas: I think the scope here should be "test". Do you agree?

punkratz312: i think it would not effect the target classed at all, but you´re right. we want 100% nothing less 👍

benas:
Hey, why did you close this PR? That was a good idea to reduce boilerplate with lombok!

punkratz312:
Sry, was a mistake - I´m not used to github. I will create a new one ^^

Issue #42, opened 2015-12-02
Failing to use jPopulator with lombok.


jsmirnov:
Hi,

In my project I am using lombok, to hide all boilerplate part of the code. I am trying to use jPopulator, but for some reason I am getting following error message:

`java.lang.NoSuchMethodException: Property '***' has no setter method in class ***`

Might be I am doing something wrong? Or these libraries currently can't be used together atm?

P.S. My problem is described here as well: https://stackoverflow.com/questions/34038236/how-to-make-benas-jpopulator-work-with-lombok

MAMABA:
Hello,

I think that private field without a setter cannot have a value with jPopulator... Have you tried to use jPopulator after compiling your class with lombok? 

Regards,

jsmirnov:
@MAMABA 
Hello,
You are right the field is private with no setter.
But I am not sure why on runtime populator doesn't see any setter, while on previous lines I am using  setters/getters generated by lombok.

jsmirnov:
@MAMABA 
It seems that I found something:
while in master version field is set as accessble:
https://github.com/benas/jPopulator/blob/master/jpopulator-core/src/main/java/io/github/benas/jpopulator/impl/PopulatorImpl.java#L260
in 1.2.0 tag field is set via `org.apache.commons.beanutils.PropertyUtils` which will fail if there is no accessor method.

I'll try master version

jsmirnov:
It seems that there is no normal way to include master version through maven. So I am confused a little bit.

MAMABA:
I remembre that someone has requested this feature....
I propose you to package and install the master via maven and include it into your POM to do a test.

benas:
Hi @jsmirnov 

Thank you for pointing out this issue.
There is no reason jPopulator may not work with lombok.

jPopulator is intended to populate standard Java Beans (with default constructor, getters and setters), see the note [here](https://github.com/benas/jPopulator/wiki/Key-APIs).

It uses apache commons beanutils behind the scene to set random values to fields. `PropertyUtils` complains when there is no setter for the field, and this is fine with what jPopulator was designed for.

To use the snapshot version, you'll need to declare the sonatype snamshot repository:

```xml
<repository>
    <id>ossrh</id>
    <url>https://oss.sonatype.org/content/repositories/snapshots</url>
</repository>
```

and include the snapshot version in your pom. I'll update the documentation with these instructions.
[...]

PR #124, opened 2016-04-25
Integrate Lombok

punkratz312:


random-beans/src/main/java/io/github/benas/randombeans/util/Constants.java
```diff
@@ -34,37 +34,34 @@
  *
  * @author Mahmoud Ben Hassine (mahmoud.benhassine@icloud.com)
  */
-public abstract class Constants {
```
PascalSchumacher: I think this should stay an abstract class (constant interfaces are a bit of an anti-patter see e.g. Effective Jave "Item 19: Use interfaces only to define types").

benas: +1 . Same here. I think it would be better to keep it a class rather than an interface

benas:
Hi @punkratz312 

Thank you for this PR!

I was thinking of using lombok to reduce boilerplate in test code, not production code.

Could you please make the dependency to lombok test scoped and keep only changes that affect test code? Many thanks upfront.

Kind regards
Mahmoud

benas:
Just a side note (not related with this issue), we've got some surprises with lombok [here](https://github.com/benas/random-beans/issues/42). The `@Accessors(chain = true)` makes lombok generate fluent setters (which is a good thing), but that breaks the java bean model and makes bean introspection fail. Interesting thread if you are interested 😉

punkratz312:
Whats your motivation not to use lombok in production as well?

benas:
There is nothing wrong with lombok for production code (it's compile time after all), I love this project! I was just thinking that you were about to add it in test code when I first saw your PR (Your PR was impacting only test code at first).

I'll do some a minor modification though. The `Constants` interface should remain as a class IMHO.

Thank you for this addition! I'll add you to the awesome contributors list :wink:

PR #128, opened 2016-05-09
Use java.lang.reflect.Array.* for primitive array "generic" operations

bedla:
Hi,
I have "solved" bounty found in source code. It is less verbose solution but it uses java.lang.reflect.Array#set method. I know that this is not the best solution (I am aware of reflection and some performance penalty of native methods) but it could be good enough ... so if you like it, merge it, or abandon it :)
Ivos


benas:
Hi Ivos,

That's fantastic! You deserve a bounty :smile:

bounty

Random Beans uses reflection anyway, so no problem with `Array#set`.

Thank you for this improvement.

Best regards
Mahmoud

bedla:
Thx ... the Bounty was delicious :)

PR #129, opened 2016-05-09
update java-faker version to 0.10

PascalSchumacher:
This version of java-faker is android compatible and fixes a bug when a seed is used in combination with `regexify`.


PascalSchumacher:
This version of java-faker is android compatible.

benas:
Good. Thanks

Issue #130, opened 2016-05-09
Range based randomized annotation


dovidkopel:
I would like to have the capacity to specify a minimum and maximum values for range based input via an annotation.

benas:
fixed in #132

PR #132, opened 2016-05-10
Range based randomized annotation #130

dovidkopel:
I came up with this. It's dynamic enough and provides the capacity to still use a single annotation. I added `common-lang3` because it has the `ConstructorUtils` (https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/reflect/ConstructorUtils.html). 

For a first stab at it. It is functional with a test as well. If you pass no arguments then it just creates the default annotation. If there are a lot of constructors with different number of arguments and types then they aren't supported right now. I'm sure that the kind of approach that is used in the `BeanFactory` would handle this more cleanly.

Let me know if you have any problems or questions.

Thanks,
DBK

### DIFF
--- a/random-beans/src/main/java/io/github/benas/randombeans/util/ReflectionUtils.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/util/ReflectionUtils.java
@@ -285,10 +285,10 @@ public class ReflectionUtils {
         return (Randomizer<T>) clazz.newInstance();
     }
 
-    private static <T> T castPrimitive(Class<T> argType, String a) {
+    public static <T> T castPrimitive(Class<T> argType, String a) {
         if(ClassUtils.isPrimitiveOrWrapper(argType)) {
             if(argType.equals(String.class)) {
-                return (T) String.valueOf(a);
+                return (T) a;
             } else if(argType.equals(Long.class)) {
                 return (T) Long.valueOf(a);
             } else if(argType.equals(Integer.class)) {
@@ -299,9 +299,11 @@ public class ReflectionUtils {
                 return (T) Double.valueOf(a);
             } else if(argType.equals(Float.class)) {
                 return (T) Float.valueOf(a);
-            } else {
-                return (T) a;
             }
+        } else if(argType.equals(byte[].class)) {
+            return (T) a.getBytes();
+        } else {
+            return (T) a;
         }
         return null;
     }
--- a/random-beans/src/main/java/io/github/benas/randombeans/util/ReflectionUtils.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/util/ReflectionUtils.java
@@ -24,7 +24,9 @@
 
 package io.github.benas.randombeans.util;
 
+import io.github.benas.randombeans.api.Randomizer;
 import lombok.experimental.UtilityClass;
+import org.apache.commons.lang3.ClassUtils;
 
 import java.lang.reflect.*;
 import java.util.*;
@@ -263,6 +265,46 @@ public class ReflectionUtils {
             }
         }
         return actualTypeArguments;
+    }
+
+    public static <T> Randomizer<T> newInstance(Class<T> clazz, String[] args) throws IllegalAccessException, InvocationTargetException, InstantiationException {
+        if(args != null && args.length > 0) {
+            for(Constructor c : clazz.getConstructors()) {
+                if(c.getParameterCount() > 0 && c.getParameterCount() == args.length) {
+                    Object[] nArgs = new Object[args.length];
+                    Class[] argTypes = c.getParameterTypes();
+                    for(int x=0; x < args.length; x++) {
+                        Class<?> argType = argTypes[x];
+                        String a = args[x];
+                        nArgs[x] = castPrimitive(argType, a);
+                    }
+                    return (Randomizer<T>) c.newInstance(nArgs);
+                }
+            }
+            return (Randomizer<T>) clazz.newInstance();
+        }
+        return null;
+    }
+
+    private static <T> T castPrimitive(Class<T> argType, String a) {
+        if(ClassUtils.isPrimitiveOrWrapper(argType)) {
+            if(argType.equals(String.class)) {
+                return (T) String.valueOf(a);
+            } else if(argType.equals(Long.class)) {
+                return (T) Long.valueOf(a);
+            } else if(argType.equals(Integer.class)) {
+                return (T) Integer.valueOf(a);
+            } else if(argType.equals(Boolean.class)) {
+                return (T) Boolean.valueOf(a);
+            } else if(argType.equals(Double.class)) {
+                return (T) Double.valueOf(a);
+            } else if(argType.equals(Float.class)) {
+                return (T) Float.valueOf(a);
+            } else {
+                return (T) a;
+            }
+        }
+        return null;
     }
 
 }
--- a/random-beans/src/main/java/io/github/benas/randombeans/util/ReflectionUtils.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/util/ReflectionUtils.java
@@ -24,6 +24,7 @@
 
 package io.github.benas.randombeans.util;
 
+import io.github.benas.randombeans.annotation.RandomizerArgument;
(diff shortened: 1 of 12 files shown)


dovidkopel:
w00t! We can make this better. Spring uses this:
https://github.com/spring-projects/spring-framework/blob/master/spring-beans/src/main/java/org/springframework/beans/factory/support/DefaultListableBeanFactory.java#L337

Which essentially adds some additional checking and tries to find the best fit constructor. We can adapt this in the future.

random-beans/src/test/java/io/github/benas/randombeans/randomizers/range/IntegerRangeRandomizerTest.java
```diff
@@ -75,4 +77,31 @@ public void shouldAlwaysGenerateTheSameValueForTheSameSeed() {
...
+            return price;
+        }
+    }
+
+    @Test
+    public void annotationShouldWorkWithRange() {
+        TestData data = null;
+        for(int x=0; x < 10; x++) {
+            data = new EnhancedRandomBuilder().build().nextObject(TestData.class);
+            then(data.getPrice()).isBetween(200, 500);
```
benas: This should be `then(data.getPrice()).isBetween(200, 220);` right?

random-beans/src/main/java/io/github/benas/randombeans/util/ReflectionUtils.java
```diff
@@ -265,4 +267,45 @@ public static boolean isWildcardType(final Type type) {
...
+    public static <T> Randomizer<T> newInstance(Class<T> clazz, String[] args) throws IllegalAccessException, InvocationTargetException, InstantiationException {
+        if(args != null && args.length > 0) {
+            for(Constructor c : clazz.getConstructors()) {
+                if(c.getParameterCount() > 0 && c.getParameterCount() == args.length) {
+                    Object[] nArgs = new Object[args.length];
+                    Class[] argTypes = c.getParameterTypes();
+                    for(int x=0; x < args.length; x++) {
+                        Class<?> argType = argTypes[x];
+                        String a = args[x];
+                        nArgs[x] = castPrimitive(argType, a);
```
benas: Do we really need to cast to primitive types? I think the method name should be just `cast` since it returns wrapper types. What do you think?

benas:
Hi Dovid,

Thank you for this PR! I've added a couple of inline comments.

I was thinking about `DateRangeRandomizer`. How can we specify date range? Since args are strings, we should specify the date format (the expected format will be documented, no pb). I see your additions does not support date types. Do you think you can add support for dates in this PR?

Kind regards
Mahmoud

dovidkopel:
Yeah that should be fine. Ultimately the code I wrote to handle the annotation parameter conversion is capable of handling any constructor as long as the type is accompanied alongside the inputted argument and its index. The class must be able to support a means to convert to that data type.

In truth we could probably make this a great deal more flexible by adding the Jackson ObjectMapper library. With that we can use the `convertValue()` and `readValue()` methods. I am going to look into a clean well designed issue for handling this.

Get back to you soon. Keep this guy open.

benas:
No problem. Take your time.

dovidkopel:
Got it! Running tests now. Super flexible. Inspired by the Spring AMQP (Thank @garyrussell) annotations patterns.

dovidkopel:
```
        @Randomizer(value=DateRangeRandomizer.class, args={
                // 2016-01-10
                @RandomizerArgument(value="1452384000000", type=Date.class),
                // 2016-01-30
                @RandomizerArgument(value="1454112000000", type=Date.class)
        })
        private Date date;
```

This is the sort of input that we can handle now. Let me know if this is what you had in mind.

benas:
This is by far more elegant! :+1: for `@RandomizerArgument`

Is the date value a timestamp from unix epcoh? Since the value is a string, why not let the user define the date with a clearly documented expected format and let the library do the conversion?

As a user, I want to input date values in a human readable format. Do you agree?

random-beans/src/test/java/io/github/benas/randombeans/randomizers/range/DateRangeRandomizerTest.java
```diff
@@ -95,4 +100,39 @@ public void whenSpecifiedMaxDateIsNull_thenShouldUseDefaultMaxValue() {
...
+            this.date = date;
+        }
+    }
+
+    @Test
+    public void annotationShouldWorkWithRange() {
+        DateRangeRandomizerTest.TestData data = null;
+        for(int x=0; x < 10; x++) {
+            data = new EnhancedRandomBuilder().build().nextObject(DateRangeRandomizerTest.TestData.class);
+            System.out.println(data.getDate().toString());
```
benas: This could be deleted safely

random-beans/src/test/java/io/github/benas/randombeans/randomizers/range/DateRangeRandomizerTest.java
```diff
@@ -95,4 +100,39 @@ public void whenSpecifiedMaxDateIsNull_thenShouldUseDefaultMaxValue() {
...
+        public void setDate(Date date) {
+            this.date = date;
+        }
+    }
+
+    @Test
+    public void annotationShouldWorkWithRange() {
+        DateRangeRandomizerTest.TestData data = null;
+        for(int x=0; x < 10; x++) {
+            data = new EnhancedRandomBuilder().build().nextObject(DateRangeRandomizerTest.TestData.class);
```
benas: You can use the shortcut static method `EnhancedRandom.random` here (since you are using all defaults):

`data = random(DateRangeRandomizerTest.TestData.class);`

dovidkopel:
Yeah so to handle that we would need a tad more intelligence. Right now I was simply using the `ObjectMapper` convert value. Which would attempt to take the `String` ...whatever its data and convert it to the passed types. For instance if you are using `JodaTime` I'm certain that the human readable format as a `String` will work since you can use that in the constructor. Give it a try. Theoretically if you want additionally flexability you can have a Factory or Converter that you can pass in that will invoke that function when needed to convert the argument into the requested type. 

We can do lots of cool stuff in the future. I think this is good for now ;)

random-beans/src/test/java/io/github/benas/randombeans/randomizers/range/DateRangeRandomizerTest.java
dovidkopel: K thanks for future reference.

benas:
> I'm certain that the human readable format as a String will work since you can use that in the constructor

Ok great. Please add a test with:

```diff
@Randomizer(value=DateRangeRandomizer.class, args={
--   // 2016-01-10
--   @RandomizerArgument(value="1452384000000", type=Date.class),
++ @RandomizerArgument(value="2016-01-10", type=Date.class),
--   // 2016-01-30
--   @RandomizerArgument(value="1454112000000", type=Date.class)
++ @RandomizerArgument(value="2016-01-30", type=Date.class)
})
private Date date;
```

and it would fine to merge your PR. I'll add documentation for the expected date format later on.

Many thanks in advance.

dovidkopel:
Anything else you waiting on from me to close this and merge it in?

random-beans/src/main/java/io/github/benas/randombeans/util/ReflectionUtils.java
```diff
@@ -24,7 +24,10 @@
 
 package io.github.benas.randombeans.util;
 
+import io.github.benas.randombeans.annotation.RandomizerArgument;
+import io.github.benas.randombeans.api.Randomizer;
 import lombok.experimental.UtilityClass;
+import org.apache.commons.lang3.ClassUtils;
```
PascalSchumacher: Look like the import is not used anymore and the commons-lang3 dependency can therefore be removed?

benas:
Hi @dovidkopel 

Didn't noticed that you've added a test with string input for dates. Thank you for this update!

As said by @PascalSchumacher , if the `ClassUtils` is no more used, I'll remove the dependency to `common-lang3`.

Kind regards
Mahmoud

PR #134, opened 2016-05-19
API sugar

punkratz312:


punkratz312:
please let my know what do you think

benas:
Hi @punkratz312 

Thank you for this PR.

> Rename API to improve readability

That's fine. What about something like:

```java
Stream<Person> persons = EnhancedRandom.randomStreamOf(100, Person.class);
```

This would be read like plain English: "random stream of 100 persons" 😉 What do you think?

> Add method randomList

That was the case [before](https://github.com/benas/random-beans/blob/83ea4a2b535fbd6d4c68ec8369589952dfc24166/random-beans/src/main/java/io/github/benas/randombeans/api/EnhancedRandom.java#L58). The API provided a method that returns a List. But many users asked why there is no method that return a Set? So I decided (since v3 with Java 8) to make the API return a stream and let the user choose to collect items in a list or set using collectors.

> Add sugar API - size is not mandatory anymore

That was [proposed in the API](https://github.com/benas/random-beans/commit/26631fc519371a445e30c2dd17e3434e32e74bdd#diff-d66b710331bc5de7b72cf5591e43ec4aL50) [too](https://github.com/benas/random-beans/commit/26631fc519371a445e30c2dd17e3434e32e74bdd#diff-0b06fc27e90c43b846cb495cb703ea4bL85). But I decided to remove it and leave it to the user.

Kind regards
Mahmoud

punkratz312:
wow, i think i made a successful "amend" - didnt thought i would made it ^^

punkratz312:
In my opinion its a clear benefit for out API to obtain direct collection types (Add sugar methods to provide direct access to collection types) - the costs aren´t that high to maintain or to extend.

I would take care of this.

benas:
Great! Thank you for the update.

Just a quick note, please avoid to add, remove and modify code at the same time. Try to keep the PR small and focused. Many thanks upfront.

benas:
Any reason why you have [removed](https://github.com/benas/random-beans/pull/134/commits/19fe18949d03ce244d8ab353f29124004edd99cf) the dependency to mockito? Looks like this breaks the build. But no problem, I've fixed that.

PR #136, opened 2016-05-28
make max collection size configurable: the default 255 sometimes is too big for bigger objects having a lot of lists

valters:
If you are randomizing a large object hierarchy, the default list size might be killing you. Sometimes you just need a few small lists here and there (such as, if you will be serializing the randomized objects as part of test).
Use the new config option as:

EnhancedRandom enhancedRandom = EnhancedRandomBuilder.aNewEnhancedRandomBuilder().maxCollectionSize( 5 ).build();

Now your randomized collections will have max 5 items. Also this ensures the collection has at least one item created - so that you can be sure some code path is not skipped.

### DIFF
--- a/random-beans/src/main/java/io/github/benas/randombeans/CollectionPopulator.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/CollectionPopulator.java
@@ -51,7 +51,7 @@ class CollectionPopulator {
 
     @SuppressWarnings({"unchecked", "rawtypes"})
     Collection<?> getRandomCollection(final Field field, final PopulatorContext context) {
-        int randomSize = enhancedRandom.nextInt(MAX_COLLECTION_SIZE);
+        int randomSize = collectionSize();
         Class<?> fieldType = field.getType();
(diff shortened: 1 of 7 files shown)


benas:
Hi Valters,

Thank you for your PR! It was possible in a previous version to configure the collection size but this has been removed in favor of registering one of the collection randomizers for the given field. Looks like this decision was not a good idea, I must admit. I invite you to look at the [discussion about this point on gitter](https://gitter.im/benas/random-beans?at=57226779961e8a522e63f40d).

Since this feature has been requested many times, I have decided to put it back and add more options at construction time to achieve something like:

```java
EnhancedRandom enhancedRandom = EnhancedRandomBuilder.aNewEnhancedRandomBuilder()
   .maxCollectionSize(10)
   .charset("UTF-8")
   .maxStringLength(50)
   .integerRange(1, 10)
   .dateRange(lastYear, nextYear)
   . // all parameters here
   .build();
```

So your PR is the first of a long serie :smile: And it is very welcome!

Thank you for this addition.

Kind regards
Mahmoud

PR #139, opened 2016-05-29
throw ObjectGenerationException when a Randomizer can not be created …

### LINKED ISSUE #131: InstantiationException when providing a randomizer without default constructor in @Randomizer
The `@Randomizer` annotation requires a randomizer with a default constructor. If no such a constructor is defined, random beans throws a `InstantiationException`. Instead, the library should throw a `ObjectGenerationException` (as declared in the API) with the `InstantiationException` as root cause

PascalSchumacher:
…(fixes #131)


benas:
No comments.. That's great! Thanks

PR #152, opened 2016-06-16
add EnhancedRandomBuilder#exclude(Class... types) (fixes #151)

### LINKED ISSUE #151: Add method for skipping randomization of types?
Although using  `EnhancedRandomBuilder#randomize(Bean.class, new SkipRandomizer())` is easy, adding a method like `EnhancedRandomBuilder#exclude(Class... types)` would make using this feature even easier and a lot more obvious.

What do you think?

PascalSchumacher:


benas:
yes, good idea. thanks for the PR!

PR #154, opened 2016-06-17
generate a empty collection if the collection has a type variable gen…

### LINKED ISSUE #153: Exception when randomizing collection with type variable generics
While looking into https://github.com/benas/random-beans/issues/150 I noticed that random beans can not randomize beans with a collection with a type variable as a generic type.

PascalSchumacher:
…eric type (fixes #153)


benas:
Indeed, this is not covered.
Thanks for the fix!

PR #162, opened 2016-06-29
Support for @Pattern; generate random strings that match regex

rmcquary:


benas:
Fantastic! Thank you Rebecca for this addition.

I was not aware of generex, cool library :smile: 

Let me merge this contribution, many thanks.

PascalSchumacher:
I guess this supersedes the existing RegularExpressionRandomizer: https://github.com/benas/random-beans/blob/4a1c01b73c77e56f977373b3b927e969d4c71bc9/random-beans-randomizers/src/main/java/io/github/benas/randombeans/randomizers/RegularExpressionRandomizer.java so the old one should be deprecated.

benas:
@PascalSchumacher I hoped you posted your comment before me ..

Quite embarrassing to have 2 regexp randomizers indeed..

There are 2 important points to consider:

* Either make the new `RegExpRandomizer` private and move it (along with its generex dependency) to the `random-beans-validation` module since it used only there. In this case, we still have two regexp randomizers ..

* Or keep the new one and deprecate the `RegularExpressionRandomizer`. But in this case, there is a point that was missed: If the `charset` param is set, the `RegExpRandomizer` must generate strings from the specified charset and looks like generex does not support this.. @rmcquary Have you an idea on how to address this?

PascalSchumacher:
>But in this case, there is a point that was missed: If the charset param is set, the RegExpRandomizer must generate strings from the specified charset and looks like generex does not support this..

Imho this is not critical, as with a regex you usually define the characters to use for the regex with things like character classes etc.

benas:
Looks like faker uses generex behind the scene:

https://github.com/DiUS/java-faker/blob/master/src/main/java/com/github/javafaker/service/FakeValuesService.java#L144

So having duplicate classes to generate random regular expressions makes no sense. I'll keep the current `RegularExpressionRandomizer` which is faker based (which in turn is generex based :smile: ) to add support for `@Pattern`. The PR is still valid and great, I'll just add the dependency to `random-beans-randomizers` to be able to use the `RegularExpressionRandomizer`

PR #165, opened 2016-07-03
update assertj-core version to 3.5.1

PascalSchumacher:


benas:
Thanks for this update. `hasNoNullFieldsOrProperties` is a gem!

Issue #164, opened 2016-07-01
Add an option to not replace value in case of null


fedotxxl:
This is related to #163
Suppose I've got `Job` class with `Set<Employee> employees = new HashSet<>();` field. If randomizer will return `null` default `HashSet` will be replaced with `null`. 

Is it possible to not replace default value in case of `null`? Something like  `DefaultValueOrNullRandomizer`

benas:
Hi,

Just to make sure I understand your requirement, you have the following class:

```java
public class Job {
     private Set<Employee> employees = new HashSet<>();
}
```

and you ask random beans to generate a random instance of `Job`. What are you expecting to get in the `employees` field? An empty set?

As of version 3.1, random beans will randomize the `employees` field and I believe this is a bug (somehow related to [inline initialization](https://github.com/benas/random-beans/blob/master/random-beans/src/test/java/io/github/benas/randombeans/FieldExclusionTest.java#L281). Inline initialization of fields should not be changed in regards to what the class designer has decided.

If I have the following class:

```java
@Data
    public static class BeanWithDefaultValues {
        Set<String> strings = new HashSet<>();
        List<Integer> integers = Arrays.asList(1, 2, 3);

        public BeanWithDefaultValues() {
            integers = Arrays.asList(1, 2, 3);
        }
    }
```

random beans will randomize both `strings` and `integers` fields and this is not what I'm expecting. The following test is failing with 3.1:

```java
@Test
public void inlineInitializationShouldNotBeChanged() {
        EnhancedRandom enhancedRandom = aNewEnhancedRandomBuilder().build();

        BeanWithDefaultValues bean = enhancedRandom.nextObject(BeanWithDefaultValues.class);

        assertThat(bean).isNotNull();
        assertThat(bean.getStrings()).isEmpty();
        assertThat(bean.getIntegers()).containsExactly(1, 2, 3);
}
```

So this should be the default behavior, not a configurable feature. Do you agree?

Kind regards
Mahmoud

PascalSchumacher:
Hi Mahmoud,

I agree. Not overwriting fields values follows the principle of least surprise and should be the default.

fedotxxl:
I always predefine `lists`/`maps` and etc (as I described before) to prevent `NullPointerException`. But I still want this fields to be randomized (as it works now). I think this is correct behavior.

>What are you expecting to get in the employees field? An empty set?
Sometimes empty set / sometimes set with random values. But not null value.

benas:
[...]

PR #166, opened 2016-07-03
add option to keep default values of fields of randomized objects (fi…

PascalSchumacher:
…xes #164)


benas:
Pascal,

Thank you for this fix. I would like to know what you think about [this](https://github.com/benas/random-beans/issues/164#issuecomment-230269797).

I think we can simply keep the `getFieldValue` method and check if the field is not null [here](https://github.com/benas/random-beans/pull/166/files#diff-58578e90e7764fe94c4b0d2547da4d1aR150). This will make inline initialization kept intact.

Cheers
Mahmoud

benas:
Thank you for your feedback! We agree, the `keepDefaultNonNullFieldValues` should be the default.

Can you update the PR please? Or I can apply changes myself if you want. I've got some free time today horaay :smile:

PascalSchumacher:
I have updated the pull request. As usual feel free to change/improve anything. :smile:

At the moment primitive fields are always randomized. I guess this should be improved to randomize only if the default value assigned by the compiler is present.

PascalSchumacher:
I have updated the pull request to only randomize primitive fields with values that are the same as the default values assigned by the compiler.

benas:
Thanks for the update. I was thinking of `overrideDefaultInitialization` as a more descriptive name than `keepDefaultNonNullFieldValues`. It would be false by default. I'll change it, don't worry about that.

PascalSchumacher:
I agree, `overrideDefaultInitialization` is a better name. :+1:

PR #167, opened 2016-07-04
update fast-classpath-scanner version to 1.10.3

PascalSchumacher:
It claims to be even faster now: https://github.com/lukehutch/fast-classpath-scanner/commit/deba683a9297021eaa8e707379a6a92f3d4bdfec


benas:
This kind of updates is always welcome. Thanks!

PR #168, opened 2016-07-04
also use Lombok for PopulatorContextStackItem and EnhancedRandomParam…

PascalSchumacher:
…eters

Just noticed that Lombok is used for other production classes, so we might use it here as well.


benas:
Cool. Less is more 😄

PR #170, opened 2016-07-05
Dependency updates

PascalSchumacher:
looks like the last fast-classpath-scanner version did not increase, but decrease performance: https://github.com/lukehutch/fast-classpath-scanner/commit/0b25528b5a7d56c888fb62e1ed01c565e78cf058


benas:
Just in time before releasing 3.2 . I was going to update all dependencies but you did it 👍 
Thanks for the update

PR #183, opened 2016-08-12
Support for method based proxies

dovidkopel:
I added in support for the user to use a byte-buddy proxy for a class instantiation instead of just `newInstance` or objenesis. As seen my unit test, the proxy leaves things open to intercept the method not just the field. I have not yet added in this support yet. Theoretically we would have a `registerRandomizer(Method method, Randomizer<?> randomizer)` type thing. I am doing this on my library right now but would soon add that functionality to RB as well.

Enjoy!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,6 +50,7 @@
         <joda-time.version>2.9.4</joda-time.version>
         <spring-context.version>4.3.1.RELEASE</spring-context.version>
         <objenesis.version>2.4</objenesis.version>
+        <byte-buddy.version>1.4.18</byte-buddy.version>
         <fast-classpath-scanner.version>1.93.0</fast-classpath-scanner.version>
         <hibernate-validator.version>5.2.4.Final</hibernate-validator.version>
         <javax.el.version>2.2.6</javax.el.version>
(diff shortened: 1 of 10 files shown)


benas:
Hi Dovid,

Thank you for this PR. I'm not sure I understand the purpose of your changes.

Can you please give an example of the feature you are adding from the user's point of view?

Thank you.
Mahmoud

dovidkopel:
The purpose is several fold. 

Not all fields/properties cannot be checked at compile-time. Methods that are accessible may be checked at compile-time for accessibility. This means that you can have more transparency when dealing with beans. My use case for this is writing a rule where a may say an object of class X when you invoke the method `getStatus()` should equal value x. In order to use this library I would need to have `getStatus()` point to a field.  We would  be setting that field with a random value. However, I may wish to have much more complex logic or not have a field at all. I may want to randomize an interface where there are no fields at all just methods. This was not possible at all with the previous solution. This approach allows for a great deal more versatility and fine tuning with what can be randomized. I will be adding conventions to allow the user to specify methods much like we do currently for fields.

dovidkopel:
Additionally, you can also change the nature of the randomization based on the user input. Different arguments would yield different outputs.

dovidkopel:
One other thing I'm going to add is a `RecursionHandler`. Enabling configuration of some recursion. For instance allow as many as _i_ levels of recursion then throw an error, or then return null, etc. Create a default handler and allow the user to configure and override that.

dovidkopel:
Can you please merge this in?

benas:
Hi Dovid,

Sorry for the late reply.

> I may want to randomize an interface where there are no fields at all just methods

This is actually out of scope of RB. The library has one precise goal: Given a **java bean**, it should generate a random instance of it. Interfaces with just methods are not java beans.

I really appreciate your effort, but I don't want to make the implementation as advanced as you propose with proxies, byte code manipulation libraries &co . You can still do it in your fork.

Kind regards
Mahmoud

Issue #184, opened 2016-08-15
Enum values are not being randomized


pmaciocia:
example:
```java
    private enum TestEnum {
        ONE, TWO, THREE, FOUR, FIVE;
    }

    private static void testEnum() {
        EnhancedRandom random = EnhancedRandomBuilder.aNewEnhancedRandomBuilder().build();

        for( int x=0; x<10; x++ ) {
            System.out.println(random.nextObject(TestEnum.class));
        }
    }
```
```
FOUR
FOUR
FOUR
FOUR
FOUR
FOUR
FOUR
FOUR
FOUR
FOUR
```

You can work around this by explicitly setting an EnumRandomizer for the enum
```
EnhancedRandom random = EnhancedRandomBuilder.aNewEnhancedRandomBuilder()
                .randomize(TestEnum.class, EnumRandomizer.aNewEnumRandomizer(TestEnum.class))
                .build();
```

From looking around the code it seems that the same seed is being reused for the EnumRandomizer each time, which would be fine if it was cached for each enum class.

pmaciocia:
this is with version 'io.github.benas:random-beans:3.2.0'

benas:
Hi,

Thank you for pointing out this issue, I was able to reproduce it.

I'll keep you informed once I publish the fix.

Kind regards
Mahmoud

benas:
Hi @pmaciocia 

The fix has been published in version 3.3.0-SNAPSHOT.

Could you please give it a try and keep me informed if it is ok for you?
Many thanks upfront.

Kind regards
Mahmoud

pmaciocia:
Hi @benas,

Just tested and looks like it's working now. 

Thanks for making this project - its been very useful to me.

Peter

benas:
Hi Peter,

Glad it helped! Thank you for reporting this issue.

Kind regards
Mahmoud

PR #190, opened 2016-08-23
#184 Reuse EnumRandomizer instances in EnhancedRandomImpl to ensure randomness.

rodriguealcazar:
Fix for #184.

### DIFF
--- a/random-beans/src/main/java/io/github/benas/randombeans/EnhancedRandomImpl.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/EnhancedRandomImpl.java
@@ -27,8 +27,11 @@ import io.github.benas.randombeans.api.*;
 import io.github.benas.randombeans.randomizers.misc.EnumRandomizer;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.Type;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


benas:
Hi,

Thank you for this PR. I've tried that same fix a few days ago (which I reverted because it was pushed by error):

https://github.com/benas/random-beans/commit/44909563c91946d47573036a89948799b7e2bfb8

But this does not really fix the problem. When a type has a field of type enum, the same instance should be always generated for the same seed. Hence, using a new seed every time an `EnumRandomizer` is returned makes some tests to fail. We need to find a solution for this case.

Kind regards
Mahmoud

rodriguealcazar:
I see, I should have obviously tried to run all test and understand the impact better.

> When a type has a field of type enum, the same instance should be always generated for the same seed

Is this just because some tests rely on this behaviour or is it by design?

benas:
No problem, don't worry.

This is by design and more general to enums. Given a type, the same instance should be generated for the same seed. This is the purpose of the seed. Here is the test case: https://github.com/benas/random-beans/blob/master/random-beans/src/test/java/io/github/benas/randombeans/EnhancedRandomImplTest.java#L266

rodriguealcazar:
I've added a map in `EnhancedRandomImpl` that keeps an instance of `EnumRandomizer` per enum type. That fixes the problem of always restarting the randomness with the same seed - root cause of this ticket - and keeps the "same seed gives same enum instance" property.

benas:
Fantastic! Thank you for this fix.

Kind regards
Mahmoud

PR #194, opened 2016-09-08
respect EnhancedRandomParameters#maxCollectionSize when populating an…

### LINKED ISSUE #191: how to limit the length of an array ?
hi:
      i'm using random-beans to generate javabean,how to configure to limit array field length?
      for example, String[] strArr, i want strArr.length less than 5.
      thanks.

PascalSchumacher:
… array (fixes #191)


benas:
Hello Pascal,

Thank you for your help!

Kind regards
Mahmoud

PR #205, opened 2016-09-30
populate fields of direct abstract superclass when populating concret…

### LINKED ISSUE #204: Instantiate Abstract Class
when i want to instantiate abstract class, instance of EnhancedRandom will create an object of concrete type but only fill properties of child class. if i put this abstract class in another class instance of container class will have object of concrete type of abstract class and all of properties is filled. i don't know it is a feature or bug. if it is feature  just properties of object in container class should be filled which belong to child class. just like instantiating an abstract class with EnhancedRandom. 
here is example:
``` 
abstract public class Vehicle {
    private int productionYear;

    public int getProductionYear() {
        return productionYear;
    }

    public void setProductionYear(int productionYear) {
        this.productionYear = productionYear;
    }
}

public class Car extends Vehicle {
    private long topSpeed;

    public long getTopSpeed() {
        return topSpeed;
    }

    public void setTopSpeed(long topSpeed) {
        this.topSpeed = topSpeed;
    }
}

public class Parking {
    private Vehicle vehicle;

    public Vehicle getVehicle() {
        return vehicle;
    }

    public void setVehicle(Vehicle vehicle) {
        this.vehicle = vehicle;
    }
}

 @Test
    public void test() {
        Car car1 = dummyCreator.nextObject(Car.class);
        Vehicle vehicle1 = dummyCreator.nextObject(Vehicle.class);
        Parking parking = dummyCreator.nextObject(Parking.class);
        Vehicle vehicle2 = parking.getVehicle();

       

    }

PascalSchumacher:
…e subclass (fixes #204)

### DIFF
--- a/random-beans/src/main/java/io/github/benas/randombeans/EnhancedRandomImpl.java
+++ b/random-beans/src/main/java/io/github/benas/randombeans/EnhancedRandomImpl.java
@@ -117,7 +117,8 @@ class EnhancedRandomImpl extends EnhancedRandom {
 
             // retrieve declared and inherited fields
             List<Field> fields = getDeclaredFields(result);
-            fields.addAll(getInheritedFields(type));
+            // we can not use type here, because with classpath scanning enabled the result can be a subtype
+            fields.addAll(getInheritedFields(result.getClass()));
 
             // populate fields with random data
             populateFields(fields, result, context);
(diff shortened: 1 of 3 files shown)


benas:
Great! thanks

Issue #207, opened 2016-10-05
Object initialization depth


sansherbina:
Hi,
For complicated objects with big amount of fields, when fields also have big amount of fields, I think, good to have any option which will control depth of initialization. For example, when set depth 2, initialized only Root object and his fields, but fields of fields are null.
By default this option might be turned off.

sansherbina:
Guys what do you think? If you are agree with it, I can contribute, this changes are important for me

PascalSchumacher:
I think this is a good idea.

benas:
Hi,

Indeed, this is a good idea. You are welcome to contribute to the project!

Kind regards
Mahmoud

benas:
@sansherbina I'm closing this issue since your PR was merged.

Many thanks!
Kind regards
Mahmoud

PR #210, opened 2016-10-25
Add ability reduce objects tree initialization depth by using option …

sansherbina:
Implementation for this feature Object initialization depth https://github.com/benas/random-beans/issues/207

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,6 +131,10 @@
             <url>https://github.com/dziga</url>
         </contributor>
         <contributor>
+            <name>Oleksandr Shcherbyna</name>
+            <url>https://github.com/sansherbina</url>
+        </contributor>
+        <contributor>
             <name>Rémi Alvergnat</name>
             <url>http://www.pragmasphere.com</url>
         </contributor>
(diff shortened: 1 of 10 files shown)


benas:
Great! Thank you for this addition :+1: 

A couple of notes:

* Is it better to rename maxInitializationDepth to maxRandomizationDepth? since we are randomizing fields (Initialization is correct, but randomization is more appropriate IMO) What do you think?

* Constants.UNLIMITED_INITIALIZATION_DEPTH should be **MAX**_RANDOMIZATION_DEPTH to be consistent with other constants. And probably set to Integer.MAX_VALUE by default instead of -1

Don't hesitate to add yourself to the contributors list! I'll merge your PR asap.
Many thanks upfront
Kind regards
Mahmoud

PS: Looking at your email and ChessAssistant repo, I guess you're a chess player :smile: Are you?
I'm also a chess junkie! catch me [on lichess](http://en.lichess.org/@/benas) and it will be a pleasure to meet you on a chessboard!

sansherbina:
I fixed all your notes. 
PS: You are right, I like play chess and play almost every evening) I created account on lichess. https://en.lichess.org/@/sansherbina It will be great to play with you)

benas:
Thank you for this update! I really appreciate your effort 👍 

See you on lichess 😄

PR #214, opened 2016-11-07
Null keys not supporter by TreeMap, ConcurrentHashMap, ConcurrentSkipListMap

sansherbina:
Null keys not supporter by TreeMap, ConcurrentHashMap, ConcurrentSkipListMap


benas:
Great! Thank you for the quick update :+1:

Issue #220, opened 2016-12-08
EnhancedRandomImpl.getRandomCollectionSize never returns maxCollectionSize


kermit-the-frog:
Hi

`return nextInt(maxCollectionSize - minCollectionSize) + minCollectionSize;`

gives only values between minCollectionSize and maxCollectionSize - 1

Kind regards,
Kermit_the_frog

PascalSchumacher:
Thanks for reporting! 👍

benas:
@kermit-the-frog Thank you for reporting this issue.

@PascalSchumacher Thank you for the quick fix!

Best regards
Mahmoud

PR #222, opened 2016-12-14
fix StringRandomizer never returns Strings of maxStringLength

kermit-the-frog:
same failure as in #220


kermit-the-frog:
Sorry, I haven't run the tests.

kermit-the-frog:
@benas Could you please explain the failed tests to me? As far as I understand they just make sure, that `StringRandomizer.getRandomValue` always returns the same value with specific `minLength, maxLength, seed`. But now I've changed the implementation of `StringRandomizer.getRandomValue` and the result for specific `minLength, maxLength, seed` has changed. Is it ok to change the `expectedValue` of these tests?

benas:
Thank you for reporting this issue and for your PR!

> Is it ok to change the expectedValue of these tests?

Yes, this is way to go.

benas:
Great! Thank you for the quick update.

Let's merge that.

Issue #234, opened 2017-02-11
RandomizerProxy.invoke will throw an exception if the Supplier passed to RandomizerProxy.asRandomizer is defined as package private or stricter


LucasAndersson:
Hi

It was a bit hard to write a good title, I'll try my best to explain the issue.

I noticed that RandomizerProxy.invoke threw a IllegalAccessException when I passed a Supplier as a lambda to EnhancedRandomBuilder.randomize. I'm not completely sure but I think the java compiler desugars lambdas to anonymous inner classes which in turn are transformed into package private inner classes. If you change RandomizerProxyTest.theRandomizerProxyShouldBehaveLikeTheSupplier to use a lambda the test will pass. I think this is because RandomizerProxyTest and RandomizerProxy are both in the same package, if you move RandomizerProxyTest to another package the exception I was talking about earlier is thrown and the test fails.

You could just pass Randomizer instead of Supplier and everything would be fine but I like the semantics of the latter when I want to return something that isn't actually random and if you already support Supplier then you might as well support them as lambdas as well.

I hope you understood my explanation, it's a bit hard and confusing to explain. I have a solution for this problem and I'm ready to make a PR if you want. I think it will be easier to understand if you read my PR.

Thanks for this great library by the way, I use it all the time.

PascalSchumacher:
Hi Lucas,

thanks for reporting.

If I understand you correctly the issue is that this:

```java
@Test
public void test() {
    EnhancedRandom random = EnhancedRandomBuilder.aNewEnhancedRandomBuilder()
        .randomize(String.class, (Supplier<String>) ()-> "test").build();

    assertThat(random.nextObject(String.class)).isEqualTo("test");
}
```

fails with:

````
io.github.benas.randombeans.api.ObjectGenerationException: Unable to generate a random instance of type class java.lang.String
	at io.github.benas.randombeans.EnhancedRandomImpl.doPopulateBean(EnhancedRandomImpl.java:128)
	at io.github.benas.randombeans.EnhancedRandomImpl.nextObject(EnhancedRandomImpl.java:77)
	at com.test.RandomizerProxyTest.test(RandomizerProxyTest.java:41)
Caused by: java.lang.reflect.UndeclaredThrowableException
	at com.sun.proxy.$Proxy4.getRandomValue(Unknown Source)
	at io.github.benas.randombeans.EnhancedRandomImpl.doPopulateBean(EnhancedRandomImpl.java:99)
	... 25 more
[...]

PR #235, opened 2017-02-13
issue #234 - RandomizerProxy should now be able to accept Supplier as lambda.

LucasAndersson:


benas:
Hi Lucas,

This is great. Thank you very much for your time and effort!

I'll merge your PR asap. Just a quick note, I think it is not necessary to make RandmomizerProxy public. I've just tested and both your test (in this PR) and the test [here](https://github.com/benas/random-beans/issues/234#issuecomment-279228003) still pass. Do you agree?

Kind regards
Mahmoud

LucasAndersson:
Hi Mahmoud

RandomizerProxyTest has to be in a separate package from RandomizerProxy, therefore RandomizerProxy can't be package private. Otherwise the tests will falsely pass even without my fix.

Br
Lucas

benas:
Lucas, sorry for the confusion, I was not clear enough in my previous comment.
Let's forget about RandomizerProxy visibility and take random beans as a black box for a moment.

If I understand correctly your bug reporting, random beans is unable to generate an instance of a given type when a custom supplier is passed as a lambda expression.
The test mentioned by @PascalSchumacher (which is an integration test) fails which version 3.5.0 .

If I take only the following modification ( + the unit test but not RandmomizerProxy visibility change) from your PR:

```diff
     @Override
      public Object invoke(final Object proxy, final Method method, final Object[] args) throws Throwable {
          if ("getRandomValue".equals(method.getName())) {
-            return target.getClass().getMethod("get").invoke(target, args);
+            Method getMethod = target.getClass().getMethod("get");
+            getMethod.setAccessible(true);
+            return getMethod.invoke(target, args);
          }
          return null;
      }
```

the test pass, which is for me enough to consider the bug as fixed. Do you see? This is what I meant by it is not necessary to make RandmomizerProxy public.

Now the question of making RandomizerProxy public or package private is another discussion (if you want, I can explain in details the decision of why it is package private)
Whether to put test classes in the same package as tested classes or not is yet another discussion.

The unit test you added in your PR is correct for me. What we can do is adding an integration test in a separate package to make sure everything is working as expected. What do you think?

LucasAndersson:
I understand your comment now, I was initially thinking to do that but I figured I would implement the test as close to the problem as possible. I just hoped that the visibility of RandomizerProxy wasn't important. Updated my PR according to your comments.

benas:
Thank you for the update 👍 
You could have kept the unit test, it was fine. I'll add it.

Let's merge this fix for now.

Many thanks!

PR #240, opened 2017-02-15
make random-beans-validation more random (fixes #237)

### LINKED ISSUE #237: Fundamental issue in random-beans-validation
I found fundamental problem in `random-beans-validation` implementation. Currently every randomizer created by the package will generate same value every time (or almost every time, see below).

Let's for example look into [`SizeAnnotationHandler`](https://github.com/benas/random-beans/blob/master/random-beans-validation/src/main/java/io/github/benas/randombeans/validation/SizeAnnotationHandler.java).

Every time randomizer is looked up for field annotated with `@Size` it will create [new instance](https://github.com/benas/random-beans/blob/master/random-beans-validation/src/main/java/io/github/benas/randombeans/validation/SizeAnnotationHandler.java#L56) of `CharacterRandomizer` providing it with **same** seed and therefore randomizer will generate **same** value every time. Or almost every time, because randomizer for string length is [not using](https://github.com/benas/random-beans/blob/master/random-beans-validation/src/main/java/io/github/benas/randombeans/validation/SizeAnnotationHandler.java#L54) seed (but it should). Other annotation handlers have same issue, except those generating constant values (for obvious reasons).

Ideal solution would be to somehow cache randomizers created for the field from annotation. But I'm not good with Java, so not sure how feasible/doable it is.

PascalSchumacher:


PascalSchumacher:
@benas What do you think?

benas:
perfect. This is what I also had in mind to fix this.

Issue #243, opened 2017-02-26
Make random-beans build on java 9


PascalSchumacher:
First we need a java 9 compatible lombok version (see: https://github.com/rzwitserloot/lombok/issues/1315), because currently on Java 9 Build 157 the build fails with:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.6.1:compile (default-compile) on project random-beans: Fatal error compiling: java.lang.ExceptionInInitializerError: Unable to make field private com.sun.tools.javac.processing.JavacProcessingEnvironment$DiscoveredProcessors com.sun.tools.javac.processing.JavacProcessingEnvironment.discoveredProcs accessible: module jdk.compiler does not "opens com.sun.tools.javac.processing" to unnamed module @90c01ab -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.6.1:compile (default-compile) on project random-beans: Fatal error compiling
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:212)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
        at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
        at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:80)
        at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:51)
        at org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:128)
        at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:307)
        at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:193)
        at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:106)
        at org.apache.maven.cli.MavenCli.execute(MavenCli.java:863)
        at org.apache.maven.cli.MavenCli.doMain(MavenCli.java:288)
        at org.apache.maven.cli.MavenCli.main(MavenCli.java:199)
[...]

PR #219, opened 2016-12-07
v2 enum randomization

korniltsev:
The problem: Generated enums of the same type are not random.
The problem is in the `io.github.benas.randombeans.EnhancedRandomImpl#randomize` . It creates              a new EnumRandomizer for every call

PR #248, opened 2017-03-24
Make random-beans 2.0.0 java6 compatible

devDavide:
Hi,
At first thank you for taking the time to read my pull request, :smile: It the first time I did one so I apologize in advance if there is something wrong with it.

I wanted to use random-beans for my project but I have to keep backwards compatibility with java 1.6  (We have planned a migration to java 1.8 ,but it is scheduled for the next year)  :-(
One of the requirement of the project I am working on, is being able to specify the seed used for the random generation. I tried to use jpopulator 1.2.0 (min Java version 1.6) but on such version was not possible to set a seed, such feature is available only since random-beans v 2.0.0 (min Java version 1.7)

Therefore I created a new branch **v2-java6-compatible** staring from the tag 2.0.0 and I adapt code in order to be compatible with java 1.6, What I did was just : 
* adapt diamond operators
* adapt try-multicatch 
* adapt  try-with-resources
* downgrade assertj.version to 1.7.1 (the last version compatible with java 1.6). All the other dependencies has been compiled with java version 1.6 or earlier. 

As an outcome, I think it could be useful to release a new version of the project, based on branch v2-java6-compatible, naming it something like random-beans 1.3.0., This will allow all the developer that still have legacy dependencies with java 1.6 to use this powerful and useful library.
What do you think about it ?

Many thanks upfront!
Kind regard.

Davide.


benas:
Hi Davide,

Thank you for your pull request! There is nothing wrong don't worry 😄 

The problem is that we're lacking time to maintain multiple versions of random beans at the same time. As I [said](https://github.com/benas/random-beans/pull/219#issuecomment-265958391) to another user, please modify, release and use your fork if you need support for Java 6. This is the beautiful part of OSS.

We are actively maintaining v3 (java 8) and we've [started working](https://github.com/benas/random-beans/issues/243) on v4 (java 9).  v1 and v2 have been archived.

Sorry for that and I hope you understand our situation.

Best regards
Mahmoud

devDavide:
Hi Mahmoud,
thank you very much for your explanation. Do not worry, I understand that maintaining multiple versions of the proyect could be complicated. So as you suggested me I will use my own fork. 
About that once I finished my work, I would like to release it and deploying on a public maven repo, using such coordinates : 
* groupId: io.github.benas
 * artifactId: random-beans
* version: 1.3.0

What is your opinion about that ? Can I use the same groupId and artifactId but different version then the original project, or it is better to change those information ?

Thank you :-)
Best regards.

Davide

benas:
> thank you very much for your explanation. Do not worry, I understand that maintaining multiple versions of the proyect could be complicated. So as you suggested me I will use my own fork.

Thank you for your comprehension.

> Can I use the same groupId and artifactId

This is not possible (unless you have my private SSH key 😉 ).

If you have access to an existing nexus or artifactory repo, you can upload your version under another groupId. If not, you can always use sonatype's nexus which is free for open source projects (see guide [here](http://central.sonatype.org/pages/ossrh-guide.html))

devDavide:
Thank you for the clear exhaustive explanation.

Have a nice day :-)

PR #250, opened 2017-03-28
Bean validation over read method

huningd:
Thanks for this great lib. It helps us really to improve our tests. Unfortunately we have some trouble with the random-beans-validation lib. Our domain model has the JSR-303 annotation on the read method and not on the fields. I know that this is questionable, but there are good points why some should prefer to annotate the read methods and not the fields. 
These changes add support to read the annotation from both field and read methods. I've created a ReflectionUtils class, which contains all the reflection handling. Now the BeanValidationAnnotationHandler and the BeanValidationRandomizerRegistry looks on the field and the read method to find the wanted bean annotation.


benas:
Hi,

Thank you for this pull request! That's fine, we can add support for introspecting annotations on getters.

A couple of notes though:

* We have already a `ReflectionUtils` class in the core module, can you please move your reflection utilities in that class? It would be great btw to add some tests for these utilities methods
* The `getAnnotationFromReadMethod` has an optional parameter. This makes the logic of the method conditional. Can we turn this parameter into non-optional and make the caller decide to call the method or not. What do you think?
* Don't hesitate to add yourself to contributors list (in readme and pom)

The rest looks great. Thank you for both test cases.
I'll merge asap and include this addition in next release.

Kind regards
Mahmoud

huningd:
Sure, I will move it to the existing ReflectionUtils and create some tests. You're right with the getAnnotationFromReadMethod was only designed as helper. I will adapt that also.

benas:
Great, please let me know when you are done.

huningd:
I've implemented your comments. Thank you for the uncomplicated contact and for adding this chances.

benas:
Fantastic! Thank you very much for all these updates.

I'll include all changes in the upcoming release. Keep tuned

benas:
Hi @huningd  

FYI, v3.7 is released and includes your changes.

Kr
Mahmoud

PR #253, opened 2017-04-28
Fix TODO in reflection utils

VanOvermeire:
This might solve the TODO in reflection utils, by creating an enum that wraps around the stuff you need in the getWrapperType enum. 

Looks a lot cleaner at any rate.


benas:
Great! it's cleaner at all levels 👍 
Thank you very much.

PR #287, opened 2017-12-01
Adding a Number Randomizer

aeneal:
I have to generate a Domain Model which uses superset Number as opposed to Int etc. I added a Number Class and reused the Integer gen to build a number. Added Tests, which pass.


benas:
Hi,

Thank you for this addition!

Kr
Mahmoud

PR #293, opened 2018-02-18
Expand MapRandomizer to support randomizing maps with the keys depend…

SButterfly:
…s on values

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,5 +208,6 @@ If you have any question, suggestion, or feedback, do not hesitate to use the [G
 * [Sam Van Overmeire](https://github.com/VanOvermeire)
 * [Valters Vingolds](https://github.com/valters)
 * [Vincent Potucek](https://github.com/punkratz312)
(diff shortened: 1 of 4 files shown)


benas:
Hi,

Thank you for this PR! However, I don't see an example of how to use this. Moreover, if we introduce a "feature" like adding the ability to generate maps where the keys depend on values, we may be asked to also add support for values that depend on keys.. So why one and not the other?

I don't think this is a common requirement to be included in the library or to modify the `MapRandomizer` class as suggested. Do you agree?

Kr,
Mahmoud

benas:
@SButterfly Any update on this?

benas:
Hi @SButterfly ,

I'm closing this PR as I don't see it as a common use case to be included in the library.

Thank you for your help anyway! If you want, you can create a repo that you own with these randomizers and I will be happy to add a reference to it in the [Extensions](https://github.com/benas/random-beans#extensions) section of the readme.

Kr,
Mahmoud

PR #307, opened 2018-07-02
Replace junit-dataprovider with junit-params.

PascalSchumacher:


benas:
That's great! Thank you Pascal for your continuous contributions 👍

PR #308, opened 2018-07-25
Migrate to fast-classpath-scanner 4.0.0

PascalSchumacher:
Not ready to merge (we should wait for the final 4.0.0 release). Just to provide some feedback for https://github.com/lukehutch/fast-classpath-scanner/issues/215.


random-beans/src/main/java/io/github/benas/randombeans/util/FastClasspathScannerFacade.java
```diff
@@ -48,8 +49,7 @@
 
     static {
         typeToConcreteSubTypes = new ConcurrentHashMap<>();
-        // disable blacklisting of JRE system jars and system packages (java.* and sun.*)
-        scanResult = new FastClasspathScanner("!!").scan();
+        scanResult = new FastClasspathScanner().unBlacklistSystemPackages().whitelistPackages("*").scan();
```
johnou: @PascalSchumacher I think you're missing either `.enableClassInfo()`  or `.enableAllInfo()`

PascalSchumacher: That was what I thought too, but `whitelistPackages` calls `enableClassInfo`.

johnou: Ah interesting, I've only skimmed over the documentation haven't integrated it yet.

lukehutch: Correct, I made `.whitelistPackages()` call `.enableClassInfo()`, since it seemed to make sense to do so -- or is it confusing to have that as the default?

However, you don't need `.whitelistPackages("*")` -- if no whitelist is provided, all packages are whitelisted for recursive scanning. I will make this more clear in the docs.

random-beans/src/main/java/io/github/benas/randombeans/util/FastClasspathScannerFacade.java
```diff
@@ -68,8 +68,9 @@
     }
 
     private static <T> List<Class<?>> searchForPublicConcreteSubTypesOf(final Class<T> type) {
-        List<String> subTypes = type.isInterface() ? scanResult.getNamesOfClassesImplementing(type) : scanResult.getNamesOfSubclassesOf(type);
-        return subTypes.stream().map(className -> {
+        String typeName = type.getCanonicalName();
+        ClassInfoList subTypes = type.isInterface() ? scanResult.getClassesImplementing(typeName) : scanResult.getSubclasses(typeName);
+        return subTypes.getNames().stream().map(className -> {
             try {
                 ClassLoader classloader = Thread.currentThread().getContextClassLoader();
```
lukehutch: You really need to not do your own classloading here, since in some situations (e.g. when scanning Spring-Boot jars on the classpath, when the scanner is not itself running within the Spring-Boot classloader), FastClasspathScanner has to create its own `URLClassLoader` to load the class. The correct way to load classes is by calling `ClassInfo#loadClass()`, or you can do it for the whole list by calling `ClassInfoList#loadClasses()` to get a `List<Class<?>>`.

lukehutch: Also, the filter line below (`.filter(Objects::nonNull).filter(currentSubType -> isPublic(currentSubType) && !(isAbstract(currentSubType)))`) should be turned into a `ClassInfoFilter` operator that filters the classes *before* classloading, not after, so that only the bare minimum number of classes is actually loaded.

PascalSchumacher:
@lukehutch Thank you very much for the review!

lukehutch:
Just a heads-up -- as of 4.0.0-beta-8, the package name of FCS has changed (removing ".lukehutch"), and the Maven groupId has changed to `io.github.fast-classpath-scanner`.

PascalSchumacher:
@lukehutch Thanks for the heads-up! I have updated the pull request to use the new groupId and the new package name.

lukehutch:
It looks like you already noticed this, based on the edited title for the PR, but the groupId and package name have changed again, since I decided to rename the whole project to ClassGraph, and that this was the best time to do it. Thanks for staying on top of this, and for testing! Things will be stable again soon :-)

PascalSchumacher:
Pull request is ready to merge now, as ClassGraph 4.0.0 has been released.

Issue #315, opened 2018-09-12
Wrong generated values when regexp owns boundary matchers (eg ^ $)


tnmtechnologies:
The random-beans-validation dependency is set.

The declaration is as follow :
```java
@JsonProperty("supportedFeatures")
@Pattern(regexp="^[A-Fa-f0-9]*$")
public String getSupportedFeatures() {
    return supportedFeatures;
}
```

The result is that the generated values always start with a ^ and ends with a $.
Example of a generated value :
` "supportedFeatures": "^d44b8dc8b6bbCd2AF56Bd$"`

This is wrong because the characters ^and $ are boundary matchers.
They are not allowed characters for the value.

benas:
Hi @tnmtechnologies 

Thank you for reporting this issue! Indeed, that's obviously incorrect.
Random Beans relies on [java-faker](https://github.com/DiUS/java-faker) to generate random strings according to a given regular expression, which in turn relies on [Generex](
https://github.com/mifmif/Generex). I think the issue is in Generex, here is a quick sample:

```java
Generex generex = new Generex("^[A-Fa-f0-9]*$");

// Generate random String
String randomStr = generex.random();
System.out.println(randomStr);
```

This sample generates `^$` which confirms the issue is in Generex. This was already reported in https://github.com/mifmif/Generex/issues/41 and apparently rejected. So we will try to see how to "sanitize" the input on random beans side so the call to faker/generex is done with the right format expected by generex.

Kr,
Mahmoud

benas:
Hi @tnmtechnologies 

We decided with @PascalSchumacher to fix the issue on RB side as it was rejected on the Generex side. The fix (in #319, many thanks to @PascalSchumacher for it) has been deployed to maven central in v3.8.0-SNAPSHOT. It would be great if you give it a try and let us know about it. I'm closing this issue for now, but feel free to add a comment here if you have any problem.

Kr,
Mahmoud

PR #319, opened 2018-11-19
RegularExpressionRandomizer: Remove leading and tailing boundary matc…

PascalSchumacher:
…hers (#315)

Not sure this is correct.

### DIFF
--- a/random-beans-randomizers/src/main/java/io/github/benas/randombeans/randomizers/RegularExpressionRandomizer.java
+++ b/random-beans-randomizers/src/main/java/io/github/benas/randombeans/randomizers/RegularExpressionRandomizer.java
@@ -76,7 +76,18 @@ public class RegularExpressionRandomizer extends FakerBasedRandomizer<String> {
 
     @Override
     public String getRandomValue() {
-        return faker.regexify(regularExpression);
+        return faker.regexify(removeLeadingAndTailingBoundaryMatchers(regularExpression));
     }
 
+    private static String removeLeadingAndTailingBoundaryMatchers(String regularExpression) {
+        String regularExpressionWithoutBoundaryMatchers = regularExpression;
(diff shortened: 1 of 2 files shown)


benas:
Hello @PascalSchumacher 

Thanks for the PR! This is how I intended to fix the problem too. However, in hindsight, should this issue be raised against faker? Faker does in turn rely on [Generex](https://github.com/mifmif/Generex) which has the issue:

```java
Generex generex = new Generex("^[A-Fa-f0-9]*$");

// Generate random String
String randomStr = generex.random();
System.out.println(randomStr);
```

This generates `^$` which is obviously not correct. Shall we raise the issue against Generex then? I'm not sure the project is maintained anymore..

Br,
Mahmoud

benas:
Yes, I will do it and see how it goes.

> Not sure this is correct.

I think it's fine as long as we support other boundary matchers as per the Javadoc of `java.util.regex.Pattern`:

```
                            Boundary matchers

^The beginning of a line
$ The end of a line
\b A word boundary
\B A non-word boundary
\A The beginning of the input
\G The end of the previous match
\Z The end of the input but for the final terminator, if any
\z The end of the input
```

This would work with a small change in your PR because those characters should appear at the beginning and the end of the regular expression. However, I'm thinking to *not* fix the issue in RB because we might need to support greedy and reluctant quantifiers which can appear in the middle of the regular expression. This would add a maintenance effort on our side while it should be on the dependency side.  Do you see the point?

And this is why I don't like dependencies!

benas:
Looks like the issue was already reported: https://github.com/mifmif/Generex/issues/41 and rejected ..

PascalSchumacher:
I would not remove more boundary matchers. One could argue for `\A`, `\z` but I would wait until this is requested to keep things simple.

As Generex will not fix this imho this should be done in random beans, but I'm o.k. if you decide otherwise.

PR #329, opened 2019-01-30
Fix regular expressions randomizer - regexp with both leading and tailing boundary matchers

weronika-redlarska:
When there are both ^ and $ boundary matchers in the regexp pattern, the tailing matcher $ is not removed. This PR fixes the issue and adds a test that is checking this case.


benas:
Fantastic! Thank you for your contribution. I added you to the contributors list (See a37a1772ff39ff47f10262c4ff97733a731306bc).

Best regards,
Mahmoud

Issue #328, opened 2019-01-30
Past/Future annotation cannot be applied to java8 date types


neetkee:
If we will put Past annotation on java 8 date type:

```java
@Past
private LocalDate localBirthDay;
```
we will get the exception:
```java
Caused by: java.lang.IllegalArgumentException: Can not set java.time.LocalDate field io.github.benas.randombeans.validation.BeanValidationAnnotatedBean.localBirthDay to java.util.Date
	at sun.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:167)
	at sun.reflect.UnsafeFieldAccessorImpl.throwSetIllegalArgumentException(UnsafeFieldAccessorImpl.java:171)
	at sun.reflect.UnsafeObjectFieldAccessorImpl.set(UnsafeObjectFieldAccessorImpl.java:81)
	at java.lang.reflect.Field.set(Field.java:764)
	at io.github.benas.randombeans.util.ReflectionUtils.setProperty(ReflectionUtils.java:96)
	at io.github.benas.randombeans.FieldPopulator.populateField(FieldPopulator.java:88)
	at io.github.benas.randombeans.EnhancedRandomImpl.populateField(EnhancedRandomImpl.java:158)
	at io.github.benas.randombeans.EnhancedRandomImpl.populateFields(EnhancedRandomImpl.java:149)
	at io.github.benas.randombeans.EnhancedRandomImpl.doPopulateBean(EnhancedRandomImpl.java:120)
	... 41 more
```

The problem is in [PastAnnotationHandler]( https://github.com/benas/random-beans/blob/master/random-beans-validation/src/main/java/io/github/benas/randombeans/validation/PastAnnotationHandler.java), I guess. It returns DateRangeRandomizer without respecting the type. Should we check for type in this class, something like this?
```java
 if (field.getType().equals(LocalDate.class)) {
            LocalDate now = LocalDate.now();
            return new LocalDateRangeRandomizer(now.minusYears(Constants.DEFAULT_DATE_RANGE), now.minusDays(1));
 }
```
Or there is a better way to do this?

PascalSchumacher:
Another option to solve this could be to use https://github.com/benas/random-beans/blob/master/random-beans/src/main/java/io/github/benas/randombeans/randomizers/registry/TimeRandomizerRegistry.java to look up the Java 8 date type.

benas:
[...]

PR #331, opened 2019-02-03
PoC: Past/Future annotation cannot be applied to Java 8 date types #328

PascalSchumacher:


benas:
Thanks for the PR! I was thinking about initial parameters (which will not be taken into account here) but that's not an issue here since we just want to generate a date in the past. And for the seed, we can get it from the constructor and use to initialize the registry.
 
My concern is that this makes a randomizer import/use a randmizer registry. I'm not sure a randomizer should import/use a registry or be aware of the registry it is registered in. A registry on the other hand can. My point is that we can live without registries, and this makes the randomizer depend on the registry API and aware of its lifecycle (needs to call `init` and pass the correct parameters, etc).

Do you agree?

PascalSchumacher:
Yes, the approach is not very elegant.

It is just a quick way to implement this feature.

I don't think there currently is a cleaner way for (higher level) randomizers to look-up (lower level) randomizers.

I do not have any problems with just closing this pull request. 😃

benas:
ok, closing this PR in favor of #335 .

PR #333, opened 2019-02-03
Make ZonedDateTime respect EnhancedRandomParameters#dateRange and #ti…

PascalSchumacher:
…meRange.

Not sure if this is the correct thing to do.

What do you think @benas ?


benas:
Yes that's correct. It is actually a bug. Some types are not randomized with respect to date/time ranges defined in parameters (like `ZonedDateTime` which you fixed here, but also `Year`, `YearMonth`, and other).

I think we should be clear in the documentation which types are concerned when we specify a date/time range through parameters.

benas:
LGTM. Thanks!

PR #337, opened 2019-02-04
Remove checks for annotation presence in bean validation randomizers …

PascalSchumacher:
…which repeat checks already done in BeanValidationRandomizerRegistry.


benas:
That's awesome! Less is more! Thank you for this.

PR #338, opened 2019-02-04
Bean validation: Add support for @PastOrPresent and @FutureOrPresent

PascalSchumacher:
Strictly speaking `PastAnnotationHandler` should be renamed to `PastOrPresentAnnotationHandler` (range: minDate (inclusive) to now (inclusive). `FutureAnnotationHandler` should be renamed to `FutureOrPresentAnnotationHandler` (range: now (inclusive) to maxDate (inclusive)).


benas:
> Strictly speaking `PastAnnotationHandler` should be renamed to `PastOrPresentAnnotationHandler` (range: minDate (inclusive) to now (inclusive).

Correct. I think we need to do something similar to `@Positive` and `PositiveOrZero`. For `@Past`, we can substract a second or two from the upper bound (now). Same for `@Future` by adding a second on the lower bound. I will take care of that. Thank you for the PR!

benas:
> I will take care of that

Done in 721a19a76b7afd33cbf0634aa7d7351a5c6e8bde

Issue #285, opened 2017-11-14
Support for meta annotations


reitzmichnicht:
Hi

Using random-beans 3.7.0 I spotted some problems with the random-beans-validator module.
This module only scans for the annotations itself, not for meta-annotations.
For example hibernate-validator contains a `@NotEmpty `annotation which in turn has a `@Size ` annotation.
This can be solved by recursion of annotations in the BeanValidationRandomizerRegistry, see Junit5 AnnotationSupport.isAnnotated() method for an example.

Not sure how to implement this without copying a lot of junit code.

Kind regards,
Michael

benas:
Hi Michael,

Valid point! Thank you for opening this issue.

I will try to make RB support meta annotations.
I'm not sure this will get released in a minor version, so I'm planning it for v4.

Kr
Mahmoud

benas:
> Not sure how to implement this without copying a lot of junit code.

That's not my first order concern, my concern is rather how to implement this on RB side. Let's imagine we manage to introspect the meta-annotation, the question is how to combine all criteria from subsequent annotations. To take the example of [`@NotNull`](https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/constraints/NotEmpty.java) from hibernate validator, it brings `@NotNull` and `@Size(min = 1)`. I wonder how to combine two (or more) randomizers to generate a random value according to this composite contract.

I've been thinking about how to tackle this, but I struggle to find a simple way to achieve it. Any idea is welcome. What do you think?

Otherwise, I will add this to the [know limitations](https://github.com/benas/random-beans/wiki/known-limitations) section of the docs and close this issue.

BTW, I noticed that RB does not support `@NotEmpty`, so I created a feature request for that (#339).

benas:
I was working on this issue the last few days and I still don't see an easy way to combine randomizers (in a generic way). Moreover, I took a look at JUnit 5 `AnnotationSupport.isAnnotated()` and I'm not ready to do something similar in here..

It is always possible to register a custom randomizer (that delegates to other randomizers) for a field annotated with the meta-annotation using `FieldDefinitionBuilder#isAnnotatedWith`.

For these reasons, I added a note about this in the [known limitations](https://github.com/benas/random-beans/wiki/known-limitations#bean-validation) section and I'm closing this issue for now.

@reitzmichnicht Thank you for this feature request anyway!

PR #351, opened 2019-03-18
@Positive and @Max

magx2:
I found out that you do not support `@Positive` and `@Max` annotations at the same time. I know that `@Positive` could be changed to `@Min(1)`, but it's definitely is not the idiomatic way.

### DIFF
--- a/easy-random-bean-validation/src/test/java/org/jeasy/random/validation/BeanValidationAnnotatedBean.java
+++ b/easy-random-bean-validation/src/test/java/org/jeasy/random/validation/BeanValidationAnnotatedBean.java
@@ -76,6 +76,10 @@ class BeanValidationAnnotatedBean {
     @Positive
     private int positive;
 
+    @Positive
+    @Max(255)
+    private int positiveWithMax;
+
     @PositiveOrZero
     private int positiveOrZero;
 
--- a/easy-random-bean-validation/src/test/java/org/jeasy/random/validation/BeanValidationMethodAnnotatedBean.java
+++ b/easy-random-bean-validation/src/test/java/org/jeasy/random/validation/BeanValidationMethodAnnotatedBean.java
@@ -42,6 +42,8 @@ class BeanValidationMethodAnnotatedBean {
 
     private int positive;
 
+    private int positiveAndMax;
+
     private int positiveOrZero;
 
     private int negative;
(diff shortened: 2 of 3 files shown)


benas:
Unfortunately it is not possible to "combine" two (or more) randomizers for a single field (in the end, a single randomizer will be selected to randomize each field). This is a known limitation documented [here](https://github.com/j-easy/easy-random/wiki/known-limitations#bean-validation).

magx2:
So maybe you need to change architecture? When user of your library use those 2 annotations he expects some behaviour. Justifying bugs with _we can't fix it_ it's poor way to resolve it.

benas:
Saying "Currently it is not possible to do X.. This is a known limitation documented here" is not justifying a bug. It is acknowledging a known issue.

If you look at [Java's know issues](https://www.oracle.com/technetwork/java/javase/8-known-issues-2157115.html), you can see for example:

> Currently it is not possible to programmatically invoke default interface methods using the JDI API. 

If you programmatically invoke default interface methods using the JDI API and it would not work as you expect, do you think Oracle folks are poorly justifying a bug with that statement?

Combining randomizers is not a bug, it is a feature that has been requested in #285. I did my best to come up with a clean solution but I could not find one. If you have an idea of how to do it, you are welcome to suggest it, I am always open to any improvement.

magx2:
Maybe add handling of `@Max` annotation to `PositiveAnnotationHandler`?

benas:
That would lead to code tangling between annotation handlers. If we follow this approach, we would also need to handle `@Min` in `NegativeAnnotationHandler` and all other combinations which is obviously not the way to go.

I should mention that Easy Random provides several way to cover your use case:

1. You can use a custom `RandomizerRegistry` where you implement the combinations you need by leveraging existing randomizers
2. You can use a custom a `RandomizerProvider` which is the main entry point for providing randomizers (this is more powerful than `RandomizerRegistry` as it give you access to the randomization context).
3. You can also register a custom randomizer for your field. This is the simplest solution IMO. Here is an example:

```java
Predicate<Field> positiveWithMaxFieldPredicate =
        FieldPredicates.isAnnotatedWith(Positive.class)
                .and(FieldPredicates.isAnnotatedWith(Max.class));
EasyRandomParameters parameters = new EasyRandomParameters()
   .randomize(positiveWithMaxFieldPredicate, new MyRandomizer());

EasyRandom easyRandom = new EasyRandom(parameters);
```

PR #358, opened 2019-06-06
Add randomizers for MAC, IPv4 and IPv6 addresses

reitzmichnicht:
Add some common randomizers provided by JavaFaker.

Also changed implementation of FakerBasedRandomizer so it can be extended from outside.


benas:
Nice additions 👍 

> Also changed implementation of FakerBasedRandomizer so it can be extended from outside.

We need to change the visibility of the field `Faker faker` in `FakerBasedRandomizer` to `protected` as well, to be able to extend the class from outside the package. Do you agree?

reitzmichnicht:
You are right, I updated the pull request

PR #360, opened 2019-06-06
Fix major performance regression.

reitzmichnicht:
Limit Size max annotation to 255 if undefined.
This is the same behavior as JPA does interpret this annotation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,7 @@ If you have any question, suggestion, or feedback, do not hesitate to use the [G
 * [Jose Manuel Prieto](https://github.com/prietopa)
 * [kermit-the-frog](https://github.com/kermit-the-frog)
 * [Lucas Andersson](https://github.com/LucasAndersson)
+* [Michael Düsterhus](https://github.com/reitzmichnicht)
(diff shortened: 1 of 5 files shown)


benas:
LGTM, thank you for the PR!

> Fix major performance regression

I'm not sure we had this size limit in a previous version, so this change cannot be qualified as a "regression" in terms of performance IMO (otherwise, I would be curious about how this performance regression was measured). It is rather a performance improvement.

In regards to tests, we need to update the bean `BeanValidationAnnotatedBean`  by adding a field:

```
@Size(min=2)
private String sizedString;
```

and updating the test `generatedValuesShouldBeValidAccordingToValidationConstraints` as well. This is for consistency with the `BeanValidationMethodAnnotatedBean` and its corresponding test.

Also please do not hesitate to add your name to the contributors list in the README, those are great contributions!

reitzmichnicht:
Yes the term regression may be misleading as it never worked better :)
But if you introduce such validation constraints in your code you see massive performance degradation of unit tests execution time.
I added the additional test.

Issue #361, opened 2019-06-07
LocalDateTimeRangeRandomizer Min-Max Exception due time of day


morvader:
Hi,

I have this data generator:

```java
EasyRandomParameters parameters = new EasyRandomParameters()
                .collectionSizeRange(nDatos, nDatos)
                .randomize(FieldPredicates.ofType(double.class),
                        new DoubleRangeRandomizer(0d, 9999d))
                .randomize(FieldPredicates.ofType(float.class),
                        new FloatRangeRandomizer(0f, 9999f))
                .randomize(FieldPredicates.ofType(int.class),
                        new IntegerRangeRandomizer(0, 9999))
                .randomize(FieldPredicates.named("endDate"), () -> LocalDateTime.now())
                .randomize(FieldPredicates.named("initDate"),
                        new LocalDateTimeRangeRandomizer(LocalDateTime.now().minusSeconds(rangoIntervalo.getSeconds())
                                , LocalDateTime.now()));
```

The issue is related to LocalDateTimeRangeRandomizer, some time the range gets yesterday, so the time of the day could be greater than now. As show is this example:

![imagen](https://user-images.githubusercontent.com/2271130/59087578-56771f80-8905-11e9-8b55-8dcd1ea3f83f.png)

That second comparator is whats is causing the exception. First LocalDateDate is actually before the second, but not the dateTimes.

It would be ok to remove the LocalTime check?

benas:
Thanks for reporting this issue! I believe it should be resolved with #373 .

You can already give it a try with `4.1.0-SNAPSHOT`. If you still have the problem, please re-open this issue.

PR #373, opened 2019-10-21
Removed fields for date and time

johkin:
The randomizer could not generate datetime if the range span midnight, since the min/max-check failed for the LocalTimeRangeRandomizer.
The new implementation breaks the current seed-based generation.

#361


benas:
That's awesome! Thank you for the PR 👍

Issue #387, opened 2020-02-03
Collections are always generated with the same size


mjureczko:
Collections are always generated with the same size, presumably due to fixed seed in EasyRandomParameters.

Failing test:
https://github.com/j-easy/easy-random/compare/master...mjureczko:collection_size_not_randomized?expand=1

benas:
Good catch, thank you for reporting this! I will check the PR #391  asap.

PR #391, opened 2020-02-18
Collections are always generated with the same size #387

mjureczko:
https://github.com/j-easy/easy-random/issues/387

The solution is based on reusing the same instance of IntegerRangeRandomizer for generating collections size. Originally, a new instance of IntegerRangeRandomizer was created whenever a collection was populated, each time with the same seed value. As a consequence IntegerRangeRandomizer always generated the same value.

Presumably, the same solution should be applied to arrays.

### DIFF
--- a/easy-random-core/src/main/java/org/jeasy/random/CollectionPopulator.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/CollectionPopulator.java
@@ -40,14 +40,17 @@ import static org.jeasy.random.util.ReflectionUtils.*;
 class CollectionPopulator {
 
     private final EasyRandom easyRandom;
+    private final IntegerRangeRandomizer integerRangeRandomizer;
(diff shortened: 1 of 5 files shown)


benas:
@mjureczko Thank you for this PR. However, I see no need to pass the parameters to the collection populator (those can be retrieved from the context which is available in the scope)  and to extract the `integerRangeRandomizer` as a field. This requires to change signature of the constructor with all required changes in tests. I also did not understand why you completely removed the `generatedObjectShouldBeAlwaysTheSameForTheSameSeed` test instead of adapting it.

Using a random seed in the collection populator is enough to fix the issue (as shown in https://github.com/j-easy/easy-random/commit/71aa6805907051d5f85724fe51bc72857304d3e4).

Thank you for raising the issue and for your PR anyway!

PR #396, opened 2020-03-07
Arrays are always generated with the same size

mjureczko:
It solves the #387 issue for arrays. I applied the same solution as @benas did in https://github.com/j-easy/easy-random/pull/391 for collections.
The change caused the `generatedObjectShouldBeAlwaysTheSameForTheSameSeed` to fail. I fixed it in the same way as in PR 391 which raised there some controversy. I extracted from the test the flow regarding arrays to a new test where I generate arrays using the same seed and then assert the equality of the generated arrays. I could adapt the original test by changing the expected array, but such a solution is not clean in my opinion. The expected array is a magic value and it doesn't manifest the test intention. It compares generated array against the magic value whereas the goal is (unless I'm mistaken) to assure that reusing the seed leads to generating the same results.

### DIFF
--- a/easy-random-core/src/main/java/org/jeasy/random/ArrayPopulator.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/ArrayPopulator.java
@@ -53,6 +53,6 @@ class ArrayPopulator {
 
     private int getRandomArraySize(EasyRandomParameters parameters) {
         EasyRandomParameters.Range<Integer> collectionSizeRange = parameters.getCollectionSizeRange();
-        return new IntegerRangeRandomizer(collectionSizeRange.getMin(), collectionSizeRange.getMax(), parameters.getSeed()).getRandomValue();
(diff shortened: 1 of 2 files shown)


benas:
Hi @mjureczko ,

Thank you for the PR! Indeed, the same issue happens with arrays.

However, the build fails with your PR. But no worries, I fixed that.

> I could adapt the original test by changing the expected array, but such a solution is not clean in my opinion. The expected array is a magic value and it doesn't manifest the test intention. It compares generated array against the magic value whereas the goal is (unless I'm mistaken) to assure that reusing the seed leads to generating the same results.

I don't agree. The arrays size does not matter in that case. Different arrays should be generated with different sizes, no matter the value of the array size.

Ensuring that the same array is always generated with the same seed is covered in another test [here](https://github.com/j-easy/easy-random/blob/ad1c8b3b061bb4dc189d2fbe363f15571506ba09/easy-random-core/src/test/java/org/jeasy/random/parameters/SeedParameterTests.java#L48), which you removed in #391 .

That said, I merged your changes and fixed the build failure. I also added you to the contributors list. Thank you for your contribution!

PR #402, opened 2020-03-26
Feature use fluent setters when populating bean

arnzel:

### DIFF
--- a/easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
@@ -137,14 +137,20 @@ public final class ReflectionUtils {
      * @throws IllegalAccessException if the property cannot be set
      */
     public static void setProperty(final Object object, final Field field, final Object value) throws IllegalAccessException, InvocationTargetException {
-        try {
-            final PropertyUtilsBean propertyUtilsBean = new PropertyUtilsBean();
-            propertyUtilsBean.addBeanIntrospector(new FluentPropertyBeanIntrospector());
-            propertyUtilsBean.addBeanIntrospector(DefaultBeanIntrospector.INSTANCE);
-            propertyUtilsBean.setProperty(object,field.getName(),value);
-        } catch ( IllegalAccessException | NoSuchMethodException e) {
-            setFieldValue(object, field, value);
+        //TODO it seems that 'propertyUtilsBean.setProperty(..)' does not work for Maps
+        if(object instanceof java.util.Map){
+            setFieldValue(object,field,value);
+        }else{
+            try {
+                final PropertyUtilsBean propertyUtilsBean = new PropertyUtilsBean();
+                propertyUtilsBean.addBeanIntrospector(new FluentPropertyBeanIntrospector());
+                propertyUtilsBean.addBeanIntrospector(DefaultBeanIntrospector.INSTANCE);
+                propertyUtilsBean.setProperty(object,field.getName(),value);
+            } catch ( IllegalAccessException | NoSuchMethodException e) {
+                setFieldValue(object, field, value);
+            }
         }
+
     }
 
     /**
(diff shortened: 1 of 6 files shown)


benas:
Thank you for this PR. The tests added here pass with the current master even without introducing the changes in `ReflectionUtils` (and adding commons-beanutils dependency). This is because in `ReflectionUtils#setProperty` method, the line:

```
PropertyDescriptor propertyDescriptor = new PropertyDescriptor(field.getName(), object.getClass());
```

fails with fluent setters (`java.beans.IntrospectionException: Method not found: setName`) and hence the field is set through reflection (in the `catch` block). This is what I tried to explain [here](https://github.com/j-easy/easy-random/issues/388#issuecomment-596790831).

Do you see?

That said, even with those fluent setters tests passing, I'm not sure if we can say ER supports fluent setters or not.

arnzel:
damn you are right. I completely forgot my orgin goal that setters are invoked while populating bean. I should add this to the test and check why the IntrospectionException is thrown. Thanks for your comment !

arnzel:
I found here an example how to use 'FluentPropertyBeanIntrospector'

https://www.javatips.net/api/org.apache.commons.beanutils.fluentpropertybeanintrospector

 I added in the test a check that setter of Bean is invoked. 

For some strange reasons it does not work for a Class which inherits from Map Class so i use the old behaviour for this kind of classes

arnzel:
@benas  could you have a look at this mr again ?

benas:
Thank you for the updates. I will test your new changes asap and get back to you.

arnzel:
@benas  any updates ?

benas:
No, unfortunately not.. I have very little spare time during these days which I'm using for another project. Easy Random is not my priority ATM. I will take a look asap. Thank you.

easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
```diff
@@ -133,6 +137,27 @@ public Object invoke(final Object proxy, final Method method, final Object[] arg
      * @throws IllegalAccessException if the property cannot be set
      */
     public static void setProperty(final Object object, final Field field, final Object value) throws IllegalAccessException, InvocationTargetException {
+        //TODO it seems that 'propertyUtilsBean.setProperty(..)' does not work for Maps
+        if(object instanceof java.util.Map){
+            setPropertyForMap(object, field, value);
+        }else{
+            try {
+                final PropertyUtilsBean propertyUtilsBean = new PropertyUtilsBean();
```
benas, line 145: No need to create a new instance for every call to `setProperty`. I will fix that on merge.

easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
```diff
@@ -133,6 +137,27 @@ public Object invoke(final Object proxy, final Method method, final Object[] arg
...
+                propertyUtilsBean.addBeanIntrospector(new FluentPropertyBeanIntrospector());
+                propertyUtilsBean.addBeanIntrospector(DefaultBeanIntrospector.INSTANCE);
```
benas, line 147: The default bean introspector is already added in the default constructor of `PropertyUtilsBean` (see https://github.com/apache/commons-beanutils/blob/commons-beanutils-1.9.4/src/main/java/org/apache/commons/beanutils/PropertyUtilsBean.java#L190). So there is no need to re-add it here.

easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
```diff
@@ -133,6 +137,27 @@ public Object invoke(final Object proxy, final Method method, final Object[] arg
      * @throws IllegalAccessException if the property cannot be set
      */
     public static void setProperty(final Object object, final Field field, final Object value) throws IllegalAccessException, InvocationTargetException {
+        //TODO it seems that 'propertyUtilsBean.setProperty(..)' does not work for Maps
```
benas, line 140: Seems to be supported for me, according to this https://github.com/apache/commons-beanutils/blob/commons-beanutils-1.9.4/src/main/java/org/apache/commons/beanutils/PropertyUtilsBean.java#L1908.

benas:
@arnzel LGTM now! rebased, squashed and merged as 12a457121c986c739a15e7354c0cc5145f17fae0. I added a couple of minor changes according to the comments above (see fec2aaa41899352b8d84f92a842177f04463df4b).

easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
arnzel: @benas  i think some related test failed but maybe it was annother problem

Issue #404, opened 2020-03-28
The same sequences of data are repeated in context of different randomizers


mjureczko:
Actual:
AbstractRandomizer creates a new instance of Random. Random is consistently initialised with the same seed, hence different instances of Random repeats the same sequences of data. I prepared a test that manifests the issue with regard to generating longs and objects containing a field of type long:
https://github.com/mjureczko/easy-random/commit/09ccd032359475e2b2d6847c1cbdd600282b7454
In the test I got exactly the same numbers from subsequent calls of nextLong() and nextObject(ClassWithLong.class) which cannot be considered random.

Expected:
nextObject(ClassWithLong.class) should not repeat the values generated by nextLong()

benas:
@mjureczko looks the same as #403 which you closed, can we close this as well?

mjureczko:
I closed #403, because I made a mistake in the test that was included there and I didn't want to keep an incorrectly reported issue open. Nonetheless, the randomness is not as random as one may expect, so this issue cannot be closed. The description of #404 is the same as #403, I only changed the test that manifests the malfunction.

benas:
ok thanks, I will take a look.

mjureczko:
I'm going to submit a merge request with a fix, but I'm not ready with it yet.

benas:
I don't see this as an issue. As explained in #406, the suggested approach breaks the design of `EasyRandom`.

You can use a custom randomizer for `ClassWithLong` and you will get different sequences of numbers. Here is a quick example based on your test (which makes it pass):

```java
@Test
void sequencesOfDataShouldNotBeReusedBetweenRandomizers() {
    //given
    class ClassWithLong {
        long number;
    }
    Set<Long> longs = new HashSet<>();
    Set<Long> objects = new HashSet<>();
    EasyRandom easyRandom = new EasyRandom(new EasyRandomParameters().randomize(ClassWithLong.class, () -> {
        ClassWithLong classWithLong = new ClassWithLong();
        classWithLong.number = new LongRandomizer().getRandomValue();
        return classWithLong;
    }));
    int N = 10;

    //when
    for (int i = 0; i < N; i++) {
        long nextLong = easyRandom.nextLong();
        longs.add(nextLong);
        long nextObject = easyRandom.nextObject(ClassWithLong.class).number;
        objects.add(nextObject);
    }

    //then
    objects.addAll(longs);
    assertThat(objects).hasSize(2 * N);
}
```

Thank you for opening this issue anyway!

[...]

PR #406, opened 2020-03-30
Sequences of data should not be reused between randomizers

mjureczko:
https://github.com/j-easy/easy-random/issues/404 - fix


benas:
Using the same seed with two `java.util.Random` instances should give the same sequence of values, as shown in the following test:

```java
@Test
void testTwoRandoms() {
    // given
    Random random1 = new Random(10L);
    Random random2 = new Random(10L);
    // when
    long long1 = random1.nextLong();
    long long2 = random2.nextLong();
    // then
    Assertions.assertThat(long1).isEqualTo(long2);
}
```

Now since `EasyRandom` is designed to be an extension of `java.util.Random` (with an additional method `nextObject`), this means I should be able to use it as a `java.util.Random` if I want to (without the additional capabilities):

```java
@Test
void testRandomAndEasyRandom() {
    // given
    Random random = new Random(10L);
    EasyRandom easyRandom = new EasyRandom(new EasyRandomParameters().seed(10L));
    // when
    long long1 = random.nextLong();
    long long2 = easyRandom.nextLong();
    // then
    Assertions.assertThat(long1).isEqualTo(long2);
}
```

The problem is that this test `testRandomAndEasyRandom` fails with this PR, which is obviously incorrect and breaks the design of `EasyRandom` in the first place. This is due to the following change:

```diff
---super.setSeed(easyRandomParameters.getSeed());
+++super.setSeed(easyRandomParameters.getSeed()+1);
```

This change modifies the user's seed which is unexpected (and might lead to a silent long overflow if the user wants to use `Long.MAX_VALUE` as seed).

I believe we are trying to solve a false problem here, as I will explain in more details in #404 .

Thank you for your PR anyway!

Issue #405, opened 2020-03-29
Parametrization of nested objects generation


mjureczko:
The behaviour regarding nested objects was described in details in https://github.com/j-easy/easy-random/issues/384. While I do not have doubts that it works as designed, I see the benefits of having the possibility to force it to work differently. I see here two scenarios:

1. Assume there is a nested object that guards its collections (fields) against null values by initializing them (e.g. in the constructor) with empty collections. As a consequence, it is not possible to have there null field unless reflection is used. However, EasyRandom uses reflection, and when reaching 'randomizationDepth' lefts the fields uninitialized. When using such objects in tests, we face NullPointerException in places where production code will never have them.
**Solution**: it would be convenient to have the possibility to chose to no to use objects with uninitialized fields and instead on the 'randomizationDepth' -1 level have empty collections.

2. When the 'objectPoolSize' is close to the number of nested objects, there is a considerable possibility of having infinity recursion, e.g. object A can have list containing among others the object A itself. This is something the production code may prevent by executing some logic, but since EasyRandom uses reflection, can easily overcome such invariants. When using such objects in tests, we may face the infinity recursion.
**Solution**:  it would be convenient to have the possibility to avoid reusing objects that are already used on the path leading to the root object, at least as long as there are in the pool objects that fulfil the condition.

benas:
> While I do not have doubts that it works as designed, I see the benefits of having the possibility to force it to work differently

Easy Random cannot cover all cases. If you want to make it work differently, you can use custom randomizers for such recursive types or provide a custom `ObjectFactory` implementation that is smart enough to covers those cases.

I'm closing this issue as explained in #412 . Thank you.

PR #412, opened 2020-04-30
#405 Parametrization of nested objects generation

mjureczko:
It only partly solves #405. It addresses only the 1st scenario described there. I find the 2nd scenario less critical (as can be avoided by increasing the size of objects pool) and more challenging to implement.

The solution introduces a new parameter that is true by default. Hence, it changes the default behaviour which can be considered as a breaking change. The defaults can be easily inverted, but I believe that it is handy to have empty collections in place of nulls.

### DIFF
--- a/easy-random-core/src/main/java/org/jeasy/random/ArrayPopulator.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/ArrayPopulator.java
@@ -42,7 +42,7 @@ class ArrayPopulator {
 
     Object getRandomArray(final Class<?> fieldType, final RandomizationContext context) {
         Class<?> componentType = fieldType.getComponentType();
-        int randomSize = getRandomArraySize(context.getParameters());
+        int randomSize = getRandomArraySize(context.getParameters(), context.useEmptyOnCurrentRandomizationDepth());
         Object result = Array.newInstance(componentType, randomSize);
         for (int i = 0; i < randomSize; i++) {
             Object randomElement = easyRandom.doPopulateBean(componentType, context);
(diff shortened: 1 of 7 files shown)


mjureczko:
I solved the 2nd scenario on a branch (https://github.com/mjureczko/easy-random/tree/%23405_scenario2). It is done on top of the 1st scenario solution.

I'm not creating a PR as I have some doubts. The solution is increasing complexity and has some performance penalty. It is controlled by a parameter and thus if not needed can be turned off, so maybe not such a big deal.

Nonetheless, the 2nd scenario cannot be solved entirely. If randomization depth is greater than the object pool size, the infinite recursion is inevitable. My change avoids infinite recursion when it is a result of the combination of randomization depth and collection size range values, i.e. when the object pool is exhausted by collections (not by randomization depth) and there is a possibility to build paths from root to leaf without repeating objects from the pool. @benas please take a look at the branch and comment what do you think about it. This PR is valid without the branch and correctly (I hope) solves scenario 1 from #405.

benas:
> Hence, it changes the default behaviour which can be considered as a breaking change.

> The solution is increasing complexity and has some performance penalty

This PR adds more parameters to make things configurable. The number of parameters start to increase significantly and as you mentioned, this increases the complexity of the code.

One of the motivation of v4 was to change the approach towards "overridable" components instead of making them configurable through additional parameters: see https://github.com/j-easy/easy-random/issues/149#issuecomment-264844185 and https://github.com/j-easy/easy-random/releases/tag/easy-random-4.0.0.RC1. So as mentioned in #405 , I believe those cases can be covered using custom randomizers or by providing a custom `ObjectFactory`.

For this reason, I'm closing this PR. I really appreciate your effort anyway!

PR #413, opened 2020-05-04
Repeatable random object on the same seed

seregamorph:
The problem: `EasyRandom` object initialized with the same `seed` parameter should generate same POJO instances. This contract came from `java.util.Random` is respected by most of `Randomizer` implementations. But there is still at least two places that generate `new Random()` object without seed. As a result the generated POJOs are different.

The context: in my project I validate equals/hashCode/toString implementations for the POJOs that have back references. I generate random beans by `EasyRandom` instance initialized with same `seed` value and then call `equals/hashCode` for such pairs, separate test calls `toString()`. Expected result: equals=true, hashCode should also equal, all three methods do not throw `StackOverflowError` (if POJO developer is not careful with equals/hashCode/toString exclude (lombok), the call will lead to infinite nested traversals).

Possible workarounds: for now it can be solved by setting `EasyRandomParameters().objectPoolSize(1)`. Because if the `populatedBeans` list collection size is single element, there is no options to randomly choose other than first.

Proposed solution: share `Random` object in `RandomizationContext`.

TDD: `RepeatableRandomTest` has test that should fail with old implementation. It's the minimal model that reproduces the issue.


seregamorph:
@benas any ideas, objections, or maybe additional clarifications required?

benas:
Hi @seregamorph , thank you for the PR! I did not look at it yet, I have very little spare time during these days which I'm using for another project. I will take a look asap and let you know. Thank you.

seregamorph:
@benas any chance it will be reviewed?

benas:
Good catch! With the same seed, we should end-up with the same generated objects (assuming equals/hashcode are correctly implemented). If this is not the case, I consider it as a bug in Easy Random. I will test your PR in details and get back to you if I need more clarification. Thank you!

seregamorph:
> With the same seed, we should end-up with the same generated objects

Exactly, this PR addresses an issue with it.

easy-random-core/src/main/java/org/jeasy/random/util/CollectionUtils.java
```diff
@@ -1,58 +0,0 @@
...
-import java.util.Random;
-
-/**
- * Collection utility methods.
- * 
- * <strong>This class is intended for internal use only.</strong>
- *
- * @author Mahmoud Ben Hassine (mahmoud.benhassine@icloud.com)
- */
-public final class CollectionUtils {
```
benas, line 36: Even if this is documented to be intended only for internal use, I do not remove public APIs without notice. Please put this back and mark it as deprecated (See how `DateUtils` has been deprecated in the same way).

easy-random-core/src/main/java/org/jeasy/random/util/CollectionUtils.java
```diff
@@ -1,58 +0,0 @@
...
-     * @return a random element from the list or null if the list is empty
-     */
-    public static <T> T randomElementOf(final List<T> list) {
-        if (list.isEmpty()) {
-            return null;
-        }
-        return list.get(nextInt(0, list.size()));
-    }
-
-    private static int nextInt(int startInclusive, int endExclusive) {
```
benas, line 55: This is actually duplicated in `RandomizationContext`. Both methods seems to be called always with `startInclusive = 0`, which mean this is no different than `Random.nextInt(bound)`, so we can get rid of these methods and use the built-in method instead (I will show how in other comments).

easy-random-core/src/main/java/org/jeasy/random/api/RandomizerContext.java
```diff
@@ -70,4 +71,12 @@
...
     EasyRandomParameters getParameters();
 
+    /**
+     * Get a random element from the list.
+     *
+     * @param list the input list
+     * @param <T>  the type of elements in the list
+     * @return a random element from the list or null if the list is empty
+     */
+    <T> T randomElementOf(List<T> list);
```
benas, line 81: In addition to being a breaking change, this does not belong to the contract of `RandomizerContext`. We can manage to have a random element from a list where this is used (ie in `FieldPopulator` and `ObjenesisObjectFactory`) without this method, see other comments.

easy-random-core/src/main/java/org/jeasy/random/FieldPopulator.java
```diff
@@ -128,7 +127,7 @@ private Object generateRandomValue(final Field field, final RandomizationContext
             value = mapPopulator.getRandomMap(field, context);
         } else {
             if (context.getParameters().isScanClasspathForConcreteTypes() && isAbstract(fieldType) && !isEnumType(fieldType) /*enums can be abstract, but can not inherit*/) {
-                Class<?> randomConcreteSubType = randomElementOf(filterSameParameterizedTypes(getPublicConcreteSubTypesOf(fieldType), fieldGenericType));
+                Class<?> randomConcreteSubType = context.randomElementOf(filterSameParameterizedTypes(getPublicConcreteSubTypesOf(fieldType), fieldGenericType));
```
benas, line 130: FieldPopulator has a reference to an EasyRandom instance (which is initialized with the same seed), so this can be replaced with:

```java
List<Class<?>> parameterizedTypes = filterSameParameterizedTypes(getPublicConcreteSubTypesOf(fieldType), fieldGenericType);
if (parameterizedTypes.size() == 0) {
    throw new ObjectCreationException("Unable to find a matching concrete subtype of type: " + fieldType);
} else {
    Class<?> randomConcreteSubType = parameterizedTypes.get(easyRandom.nextInt(parameterizedTypes.size()));
    value = easyRandom.doPopulateBean(randomConcreteSubType, context);
}
```

easy-random-core/src/main/java/org/jeasy/random/ObjenesisObjectFactory.java
```diff
@@ -47,7 +46,7 @@
     @Override
     public <T> T createInstance(Class<T> type, RandomizerContext context) {
         if (context.getParameters().isScanClasspathForConcreteTypes() && isAbstract(type)) {
-            Class<?> randomConcreteSubType = randomElementOf(getPublicConcreteSubTypesOf((type)));
+            Class<?> randomConcreteSubType = context.randomElementOf(getPublicConcreteSubTypesOf((type)));
```
benas, line 49: We can use a `Random` instance here and initialize it  (only once) with the seed from the context:

```java
    private Random random;

    @Override
    public <T> T createInstance(Class<T> type, RandomizerContext context) {
        if (random == null) {
            random = new Random(context.getParameters().getSeed());
        }
        if (context.getParameters().isScanClasspathForConcreteTypes() && isAbstract(type)) {
            List<Class<?>> publicConcreteSubTypes = getPublicConcreteSubTypesOf(type);
            if (publicConcreteSubTypes.size() == 0) {
                throw new InstantiationError("Unable to find a matching concrete subtype of type: " + type + " in the classpath");
            } else {
                Class<?> randomConcreteSubType = publicConcreteSubTypes.get(random.nextInt(publicConcreteSubTypes.size()));
                return (T) createNewInstance(randomConcreteSubType);
            }
        } else {
            try {
                return createNewInstance(type);
            } catch (Error e) {
                throw new ObjectCreationException("Unable to create an instance of type: " + type, e);
            }
        }
    }
```

easy-random-core/src/main/java/org/jeasy/random/RandomizationContext.java
```diff
@@ -105,7 +108,7 @@ boolean hasExceededRandomizationDepth() {
     }
 
     private int nextInt(int startInclusive, int endExclusive) {
```
benas, line 110: This method can be removed and replace the line where it is called with:

```java
int randomIndex = actualPoolSize > 1 ? random.nextInt(actualPoolSize) : 0;
```

easy-random-core/src/main/java/org/jeasy/random/RandomizationContext.java
```diff
@@ -148,4 +151,12 @@ public Object getRootObject() {
     public EasyRandomParameters getParameters() {
         return parameters;
     }
+
+    @Override
+    public <T> T randomElementOf(List<T> list) {
```
benas, line 156: This should be removed as explained below.

benas (changes requested):

easy-random-core/src/main/java/org/jeasy/random/util/CollectionUtils.java
seregamorph: reverted and marked deprecated

easy-random-core/src/main/java/org/jeasy/random/api/RandomizerContext.java
seregamorph: removed

easy-random-core/src/main/java/org/jeasy/random/FieldPopulator.java
seregamorph: thanks for exact code sample. done

easy-random-core/src/main/java/org/jeasy/random/ObjenesisObjectFactory.java
seregamorph: done

easy-random-core/src/main/java/org/jeasy/random/RandomizationContext.java
seregamorph: done

easy-random-core/src/main/java/org/jeasy/random/RandomizationContext.java
seregamorph: removed

benas:
Perfect! Thank you for these updates. Rebased, squashed and merged as 15e5a82e8ac58dea97aa340cad83252184e971e8. Thank you for your contribution 👍

Issue #420, opened 2020-06-07
Number-based validation annotations are not supported well on primitive wrapper


dadiyang:
Suppose we have these Long field annotated with Postive/PositiveOrZero/Negative/NegativeOrZero: 
![image](https://user-images.githubusercontent.com/28141217/83958264-4f7d4700-a8a2-11ea-957a-23c37be8fc88.png)

EasyRandom will occur an exception, because it tries to set a Long field with int value.

![image](https://user-images.githubusercontent.com/28141217/83958119-b7329280-a8a0-11ea-8b4e-27f0c00689d8.png)

onacit:
I'm facing the very same problem.
EasyRandom seems trying to set primitive values to those wrapper types.

dadiyang:
It's a general problem. And I fix it on my pull request. Hope @benas fix this problem as quickly as possible. Thanks!

lancedfr:
I have the same issue when using

`  @Positive
  private BigInteger number;
`

> org.jeasy.random.ObjectCreationException: Unable to create a random instance of type class com.SupplierTransaction
Caused by: java.lang.IllegalArgumentException: argument type mismatch
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at org.jeasy.random.util.ReflectionUtils.setProperty(ReflectionUtils.java:136)
	at org.jeasy.random.FieldPopulator.populateField(FieldPopulator.java:93)
	at org.jeasy.random.EasyRandom.populateField(EasyRandom.java:206)
	at org.jeasy.random.EasyRandom.populateFields(EasyRandom.java:195)
	at org.jeasy.random.EasyRandom.doPopulateBean(EasyRandom.java:162)
	... 65 more

When investigating this in my case it seems like
org.jeasy.random.RegistriesRandomizerProvider#getRandomizer method is returning IntegerRangeRandomizer and not BigIntegerRandomizer causing the argument type mismatch

lancedfr:
> It's a general problem. And I fix it on my pull request. Hope @benas fix this problem as quickly as possible. Thanks!

[...]

PR #421, opened 2020-06-07
Make number-based validation annotation handlers support primitive wr…

dadiyang:
Make number-based validation annotation handlers support primitive wrappers such as Long, Integer, Byte etc.
Please check #420 for the bug detail.


benas:
LGTM. Rebased, squashed and merged as 8c978032cd1462ca84180eed9f33667028aa0387. Thank you for your contribution!

Issue #425, opened 2020-07-30
Support type variables


ollik1:
Consider the following example:
```
abstract class Base<T> {
    T t;
}
class Concrete extends Base<String> {}
```
Currently `easyRandom.nextObject(Concrete.class)` would generate `t` as `java.lang.Object` which will then cause a runtime exception when used. The actual types are preserved in classes so that could be used to pick the correct randomizer instead.

sin-dm:
Hello! What are the deadline for closing this issue? 
Now I have this troubles when generating child objects. This issue must solve my problem

benas:
Duplicate/Similar to #375 , #242 and #386.  

@ollik1 I would like to get your feedback on https://github.com/j-easy/easy-random/pull/426#issuecomment-711423863 before closing this as duplicate.

seregamorph:
@benas if you have time you can check the intermediate state.
To simplify your review - it is separated into two parts:
First pivot - https://github.com/j-easy/easy-random/compare/master...seregamorph:generic-base-class?expand=1
it is a fast-forward of #426 (currently outdated) and latest `origin/master`. The build of this branch is success (note: it does not have new test cases).
Second - the fix of #426 that now supports intermediate classes to resolve generic types (new tests added). The logic is based on spring-framework spring-core `org.springframework.core.ResolvableType.forField(Field, Class)` method. All related logic copied from spring project (making spring-core dependency looks weird even if shade included), then step-by-step all unused methods were made private or deleted. You can assume that there is too many new lines, but you can simplify your review this way: compare original spring source and copied class. The order of methods is the same, only caches were removed.
The good thing about this implementation: it is promising. Please check the test `ResolvableTypeTest`, it covers the second case that you described (but still the implementation of random is incomplete) - generic collections of fields. I need a few more days, but there is a chance that I can handle it. Just wanted to know your opinion about the approach.

seregamorph:
The implementation: https://github.com/seregamorph/easy-random/compare/generic-base-class...seregamorph:generic-base-class-spring-core-no-spring?expand=1 (it's too early to raise the PR, so the delta link)

seregamorph:
One more implementation based on fasterxml-classmate dependency https://github.com/seregamorph/easy-random/compare/generic-base-class...seregamorph:generic-base-class-classmate?expand=1
[...]

Issue #386, opened 2020-01-30
Unable to create a random instance (Generic Types)


messaoudi-mounir:
The test bellow fail 

```
public abstract class ReferencedObject {
 private UUID uuid;
 public ReferencedObject(){
  }
}

public class A extends ReferencedObject  {
   public A(){}
   // some code here 
}
public class B extends ReferencedObject  {
    public B(){}
   // some code here 
}

public class DataObjectReference<T extends ReferencedObject> {
  private String contentType;
  private T referencedObject;

  public DataObjectReference() {
  }
  public DataObjectReference(T referencedObject) {
    this.referencedObject = referencedObject;
  }
}

public class C {
     private DataObjectReference<A> parent;
     public C(){}
    // some code here 
}


EasyRandom easyRandom = new EasyRandom();
C pojo = easyRandom.nextObject(C.class);

}

```
// error message

 Unable to create a random instance of type class ReferencedObject

benas:
Thank you for opening this issue. This is a duplicate of #375 . This is a [known limitation](https://github.com/j-easy/easy-random/wiki/known-limitations#type-erasure) and is not supported due to type erasure.

You need to use a custom randomizer for the generic type.

Issue #375, opened 2019-10-30
Support for any generic type


ivangreene:
EasyRandom version: `easy-random-core-4.0.0`

When inside a class, we can randomize a field such as List<T>. However this only seems to work on a limited set of known types (List, etc) and not on simply any type with parameters.

Failing test:
```java
    static class Element<T> {
        T value;
    }

    static class Container {
        Element<String> element;
    }

    @Test
    public void shouldWork() {
        EasyRandom easyRandom = new EasyRandom();
        Container container = easyRandom.nextObject(Container.class);
        assert container.element.value instanceof String;
    }
```

Here is a pretty minimal way to get the actual type:
```java
        Field elementField = Container.class.getDeclaredField("element");
        Type actualParameter = null;
        Type genericType = elementField.getGenericType();
        if (genericType instanceof ParameterizedType) {
            actualParameter = ((ParameterizedType) genericType).getActualTypeArguments()[0];
        }
        assert actualParameter == String.class;

```

benas:
Thanks for opening this issue, we already use this kind of tricks in various places of the code base.

However, we had the discussion before about supporting generic types but we decided to not support them (please read the details in #242 and #313).

ivangreene:
@benas would you be prepared to review a pull request that implements handling for these types of situations? It really limits our ability to generate random beans for many common uses of parameterized types

benas:
Why is this limiting? You can register a custom randomizer for that generic type as suggested [here](https://github.com/j-easy/easy-random/wiki/known-limitations#type-erasure).

There are a lot of cases to cover to correctly implement this feature:

* what if T in your example resolves to `Element<List<String>> element;` or `Element<Map<Integer, List<String>>> element;`?
* what if the generic type you are looking for is not defined in the root type (as explained [here](https://github.com/j-easy/easy-random/issues/242#issuecomment-441480572))
* etc

I believe the clean way to deal with this is to have something like `Field.isGenericType()` and then if true do `Field.getActualType()`, but this is not possible until reified generics are added to Java. Any other solution will be fighting type erasure with some kind of hacks IMO.  I'm of course open for a PR you are welcome! But I'm not ready to maintain any complex logic (like the one proposed in #354, btw I invite you to read the comments there) unless you really come up with something easy 😉 .

beluchin:
I put together a random value generator that fully supports generics. Consider taking a look at:
https://github.com/beluchin/rvg

Issue #242, opened 2017-02-23
Generation fails for generic enums


reitzmichnicht:
Here is a unit test for the error:
```
	@Test
	public void genericEnumTest() {
		FooBar dto = EnhancedRandom.random(FooBar.class);
	}

	public class BarDto<T extends Enum<T>> implements Serializable {
		private T status;

		public T getStatus() {
			return status;
		}

		public void setStatus(T status) {
			this.status = status;
		}
	}
	
	public class FooBar extends BarDto<FooEnum> {
	}
	
	public enum FooEnum {
		DEFAULT
	}
```

The test fails with this message:
io.github.benas.randombeans.api.ObjectGenerationException: Unable to generate a random instance of type class $FooBar
...
Caused by: io.github.benas.randombeans.api.ObjectGenerationException: Unable to create type: java.lang.Enum for field: status of class: $FooBar
...

PascalSchumacher:
Thanks for reporting.

I believe due to type erasure it is not possible to figure out what type should be stored in the `status` field. 

I would love to be proven wrong. :)

PascalSchumacher:
Of course you can help random-beans figure out what to do. For example with:

```java
@Test
public void genericEnumTest() {
  EnhancedRandom enhancedRandom = EnhancedRandomBuilder.aNewEnhancedRandomBuilder()
      .randomize(new FieldDefinition("status", Enum.class, BarDto.class), new EnumRandomizer(FooEnum.class)).build();
  FooBar dto = enhancedRandom.nextObject(FooBar.class);
}
```

reitzmichnicht:
Hi Pascal,

thanks for your answers and the example. This is a part where java really sucks compared to other generic concepts like c#
I'm wondering why this also doesn't work:
`	public class FooBar extends BarDto<FooEnum> {
		@Override
		public FooEnum getStatus() {
			return super.getStatus();
		}

		@Override
		public void setStatus(FooEnum status) {
			super.setStatus(status);
		}
	}`

Btw: Great library that is a good java counterpart to the famous C# AutoFixture

PascalSchumacher:
Hi reitzmichnicht,

thanks for your kind words. :-)

Random-Beans does work directly on fields (by-passing setters/getters).  This means for Random-Beans your new example is the same as the old one (it still can not identify the actual type of the status field).

I think reified generics are on the roadmap for java 10.

benas:
Hi @reitzmichnicht 

[...]

PR #426, opened 2020-07-30
#425 Support type variables

ollik1:
See https://github.com/j-easy/easy-random/issues/425

Pass the actual type along with a field that will be used to pick
correct randomizer. Consider the following example:
```
abstract class Base<T> {
	T t;
}
class Concrete extends Base<String> {}
```
If we only rely on `Field.getType()`, property `Concrete.t` would be
randomized as `Object` and then cause a runtime exception when used.

Java reflection API provides a way to get the actual type, in this case
`String`. This information is passed in `GenericField` so that the
correct randomizer can be picked later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,7 @@ If you have any question, suggestion, or feedback, do not hesitate to use the [G
 * [Nikola Milivojevic](https://github.com/dziga)
 * [nrenzoni](https://github.com/nrenzoni)
 * [Oleksandr Shcherbyna](https://github.com/sansherbina)
+* [Olli Kuonanoja](https://github.com/ollik1)
 * [Petromir Dzhunev](https://github.com/petromir)
 * [Rebecca McQuary](https://github.com/rmcquary)
 * [Rémi Alvergnat](http://www.pragmasphere.com)
--- a/easy-random-core/src/main/java/org/jeasy/random/EasyRandom.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/EasyRandom.java
@@ -25,6 +25,7 @@ package org.jeasy.random;
 
 import org.jeasy.random.api.*;
 import org.jeasy.random.randomizers.misc.EnumRandomizer;
+import org.jeasy.random.util.GenericField;
 import org.jeasy.random.util.ReflectionUtils;
 
 import java.lang.reflect.Field;
@@ -148,7 +149,7 @@ public class EasyRandom extends Random {
             context.addPopulatedBean(type, result);
 
             // retrieve declared and inherited fields
-            List<Field> fields = getDeclaredFields(result);
+            List<GenericField> fields = getDeclaredFields(result);
             // we can not use type here, because with classpath scanning enabled the result can be a subtype
             fields.addAll(getInheritedFields(result.getClass()));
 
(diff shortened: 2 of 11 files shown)


benas:
Thank you for opening this PR. The following tests fail with `java.lang.ClassCastException: java.lang.Object cannot be cast to java.util.List`:

```java
@Test
void genericBaseClassWithList() {
    class Concrete extends GenericBaseClass<List<String>> {
        private final String y;

        public Concrete(List<String> x, String y) {
            super(x);
            this.y = y;
        }

        public String getY() {
            return y;
        }
    }

    Concrete concrete = easyRandom.nextObject(Concrete.class);
    assertThat(concrete.getX()).isInstanceOf(List.class);
    concrete.getX().forEach(System.out::println);
    assertThat(concrete.getY().getClass()).isEqualTo(String.class);
}

@Test
void testMultipleGenericLevels() {
    abstract class BaseClass<T> {
        protected T x;
        BaseClass(T x) {
            this.x = x;
        }
        public T getX() {
            return x;
        }
    }

    abstract class GenericBaseClass<T, P> extends BaseClass<T> {
        protected P y;
        GenericBaseClass(T x, P y) {
            super(x);
            this.y = y;
        }
        public P getY() {
            return y;
        }
    }

    class Concrete extends GenericBaseClass<String, Long> {
        Concrete(String x, Long y) {
            super(x, y);
        }
    }

    Concrete concrete = easyRandom.nextObject(Concrete.class);
    assertThat(concrete.getX()).isInstanceOf(String.class);
    assertThat(concrete.getY()).isInstanceOf(Long.class);
    System.out.println("concrete.getX() = " + concrete.getX());
    System.out.println("concrete.getY() = " + concrete.getY());
}
```

This enhancement is really tricky. We had several attempts to implement it but we came to the conclusion that there is no clean way to do it until reified generics are supported in the Java reflection API (I invite you to read the discussions in #375 , #242 and #386 ). Do you agree?

That's why we ended up documenting this in the [known limitations](https://github.com/j-easy/easy-random/wiki/known-limitations#type-erasure) section.

seregamorph:
@benas for sure you are right regarding given samples. But this PR still enhances the situation, it may become a step forward.
I believe, the Pareto rule works here: such change will cover 80% of possible cases.
It covers all mine (because right now I need to workaround it like set up randomization of base `Serializable` (uppermost generic type in my case) type to one of subtypes.
The change looks harmless (had to resolve merge conflicts locally), all existing use cases still work fine. To play more safe, the introduced class `GenericField` may become package visible (instead of public), new utility methods should be placed in `org.jeasy.random` (not `org.jeasy.random.util`). This way you can keep space for breaking possible back compatibility (on implementation; not behavioral).

Anyway, that's just my humble opinion.

benas:
@seregamorph Thank you for sharing your thoughts on this!

Up until now, I was reluctant to introduce a half-backed feature in the library most of the time. But based on the popularity of this feature request, I think it makes sense to add it at least for common use cases, we can always document non supported use cases in the "known limitations" section of the docs.

That said, I would not bother with the `genericBaseClassWithList` test failing with this PR as this is similar to what is already documented as unsupported (see https://github.com/j-easy/easy-random/wiki/known-limitations#composite-collections). However, `testMultipleGenericLevels` should really pass as this a very common use case.

My other concern with this PR is the introduction of this new `GenericField` concept. This abstraction is confusing to me, especially when I have to think about when I should write `field.getField().getType()` vs `field.getType()`. Moreover, this impacts almost everywhere where `Field` is used which should now be `GenericField` (including public APIs of `ReflectionUtils` which is a breaking change). Making this concept package private and moving utility methods in `org.jeasy.random` means we would have scattered utility classes (`SomeInternalUtils` besides `ReflectionUtils`) which is not the goal in the first place.

I believe we can support the suggested use cases with a simpler approach: whenever we populate a field, we could check it if it is generic and try to get the type from its base class (or one from its hierarchy). I tried to implement this in https://github.com/j-easy/easy-random/commit/0cadecef93abbdb34f535a0572b60e4efdb28cfa. This approach has a minimal impact (only `FieldPopulator`), passes all tests of this PR + the above `testMultipleGenericLevels`, and does not introduce any regression or API breaking change.

@ollik1 @seregamorph I would be grateful if you could take a look at this attempt in order to move this feature request forward and include it in the upcoming 4.3. Many thanks upfront.

seregamorph:
@benas I will take care of if. Already have some prototypes (based on spring-core and fasterxml-classmate just for investigations). Trying to extract working prototype without new external dependencies.

benas:
@seregamorph ok thank you! I opened #439 to make it easier to review the changes. @ollik1 your feedback is welcome as well!

benas:
> I believe, the Pareto rule works here: such change will cover 80% of possible cases.

I agree, and this is a good start! Both this #426 and #439 cover those cases. But as explained [previously](https://github.com/j-easy/easy-random/pull/426#issuecomment-722041902), I had some concerns with this PR. So I'm closing it in favor of #439.

@ollik1 Thank you for your contribution anyway! And thanks to @seregamorph for the review and follow up!

PR #432, opened 2020-10-14
Pass properly updated context to a ContextAwareRandomizer

unconditional:
The stack of the context passed to any `ContextAwareRandomizer` via `setRandomizerContext()` does not have the actual currently processed field on top, but the enclosing/parent field instead. 
This makes it impossible in a randomizer to acquire and use the current field metadata to generate a random value (e.g. based on attributes of a custom annotation on it).


benas:
LGTM. Rebased and merged as 729420b14d447e33570833e5a23bc5ae9411406e. Thank you for your contribution!

PR #438, opened 2020-11-03
Fix #437: chain setters and overriden fields issue

### LINKED ISSUE #437: Regression: Unable to create a random instance (brother subclasses)
Consider this snippet:
```java
import lombok.Data;
import lombok.experimental.Accessors;
import org.jeasy.random.EasyRandom;
import org.jeasy.random.EasyRandomParameters;

public class EasyRandomTest {

    public static void main(String[] args) {
        EasyRandom easyRandom = new EasyRandom(new EasyRandomParameters());

        System.out.println(easyRandom.nextObject(BulkOperationObjectError.class));
    }

    @Data
    @Accessors(chain = true)
    public static class ObjectError {
        private String objectName;
    }

    @Data
    @Accessors(chain = true)
    public static class BulkOperationObjectError extends ObjectError {
        private FieldError fieldError;
    }

    @Data
    @Accessors(chain = true)
    public static class FieldError extends ObjectError {
        @Override
        public FieldError setObjectName(String objectName) {
            super.setObjectName(objectName);
            return this;
        }
    }
}
```

I minimized the code as much as possible. In `4.2.0` it successfully creates a random instance. In `4.3.0-SNAPSHOT` it fails with exception:
```
Exception in thread "main" org.jeasy.random.ObjectCreationException: Unable to create a random instance of type class com.trilogy.restapi.errors.EasyRandomTest$BulkOperationObjectError
	at org.jeasy.random.EasyRandom.doPopulateBean(EasyRandom.java:172)
	at org.jeasy.random.EasyRandom.nextObject(EasyRandom.java:100)
	at com.XXX.restapi.errors.EasyRandomTest.main(EasyRandomTest.java:14)
Caused by: java.lang.IllegalArgumentException: com.trilogy.restapi.errors.EasyRandomTest$BulkOperationObjectError is not assignable from com.XXX.restapi.errors.EasyRandomTest$FieldError
	at org.apache.commons.beanutils.MethodUtils.getAccessibleMethod(MethodUtils.java:793)
	at org.apache.commons.beanutils.PropertyUtilsBean.getWriteMethod(PropertyUtilsBean.java:1319)
	at org.apache.commons.beanutils.PropertyUtilsBean.setSimpleProperty(PropertyUtilsBean.java:2094)
	at org.apache.commons.beanutils.PropertyUtilsBean.setNestedProperty(PropertyUtilsBean.java:1915)
	at org.apache.commons.beanutils.PropertyUtilsBean.setProperty(PropertyUtilsBean.java:2022)
	at org.jeasy.random.util.ReflectionUtils.setProperty(ReflectionUtils.java:153)
	at org.jeasy.random.FieldPopulator.populateField(FieldPopulator.java:102)
	at org.jeasy.random.EasyRandom.populateField(EasyRandom.java:209)
	at org.jeasy.random.EasyRandom.populateFields(EasyRandom.java:198)
	at org.jeasy.random.EasyRandom.doPopulateBean(EasyRandom.java:165)
	... 2 more
```

I did not investigate it deeper yet, but for sure it may become a problem in case of version upgrade.

seregamorph:
There are two regression defects described in #437. The first: if there are two subclasses of a base class, the corrupted `PropertyDescriptor` information may be cached and reused. The second: if the base class has a field with chained accessors and a subclass overrides this field (not a setter, but a field), the randomization of an instance fails. Both cases are covered with tests:
- `EasyRandomChainedSettersTest`
- `EasyRandomOverridenFieldTest`

Note: both tests are success on version `4.2.0`, but will fail if you revert all other changes in `src/main/java` and `pom.xml`, hence it is a regression.

Also this PR removes `commons-beanutils` dependency as it is not required anymore. Please note, that there is a critical issue https://issues.apache.org/jira/browse/BEANUTILS-541 that caused one of described problems.

### DIFF
--- a/easy-random-core/pom.xml
+++ b/easy-random-core/pom.xml
@@ -58,10 +58,6 @@
             <artifactId>classgraph</artifactId>
         </dependency>
         <dependency>
-            <groupId>commons-beanutils</groupId>
-            <artifactId>commons-beanutils</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter</artifactId>
             <scope>test</scope>
--- a/easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
@@ -23,15 +23,11 @@
  */
 package org.jeasy.random.util;
 
-import org.apache.commons.beanutils.FluentPropertyBeanIntrospector;
-import org.apache.commons.beanutils.PropertyUtilsBean;
 import org.jeasy.random.annotation.RandomizerArgument;
 import org.jeasy.random.ObjectCreationException;
 import org.jeasy.random.api.Randomizer;
 import org.objenesis.ObjenesisStd;
 
-import java.beans.IntrospectionException;
-import java.beans.PropertyDescriptor;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.*;
 import java.math.BigDecimal;
@@ -59,11 +55,6 @@ import static java.util.stream.Collectors.toList;
  * @author Mahmoud Ben Hassine (mahmoud.benhassine@icloud.com)
  */
 public final class ReflectionUtils {
-
-    private static final PropertyUtilsBean PROPERTY_UTILS_BEAN = new PropertyUtilsBean();
-    static {
-        PROPERTY_UTILS_BEAN.addBeanIntrospector(new FluentPropertyBeanIntrospector());
-    }
 
     private ReflectionUtils() {
     }
@@ -140,21 +131,16 @@ public final class ReflectionUtils {
      * @throws IllegalAccessException if the property cannot be set
      */
     public static void setProperty(final Object object, final Field field, final Object value) throws IllegalAccessException, InvocationTargetException {
-        try { // to call regular setter
-            PropertyDescriptor propertyDescriptor = new PropertyDescriptor(field.getName(), object.getClass());
-            Method setter = propertyDescriptor.getWriteMethod();
+        try {
+            Method setter = getSetter(object.getClass(), field.getName(), field.getType());
             if (setter != null) {
                 setter.invoke(object, value);
             } else {
                 setFieldValue(object, field, value);
             }
-        } catch (IntrospectionException | IllegalAccessException e) {
-            try { // to call fluent setter
-                PROPERTY_UTILS_BEAN.setProperty(object, field.getName(), value);
-            } catch (NoSuchMethodException noSuchMethodException) {
-                // otherwise, set field using reflection
-                setFieldValue(object, field, value);
-            }
+        } catch (IllegalAccessException e) {
+            // otherwise, set field using reflection
+            setFieldValue(object, field, value);
         }
     }
 
@@ -607,6 +593,27 @@ public final class ReflectionUtils {
             return (Randomizer<T>) type.newInstance();
         } catch (IllegalAccessException | InvocationTargetException | InstantiationException e) {
             throw new ObjectCreationException(format("Could not create Randomizer of type: %s with constructor arguments: %s", type, Arrays.toString(randomizerArguments)), e);
+        }
+    }
+
+    private static Method getSetter(Class<?> beanType, String propertyName, Class<?> propertyType) {
+        String setterName = "set" + propertyName.substring(0, 1).toUpperCase(ENGLISH) + propertyName.substring(1);
+        // implementation note: we go from subclass to super-class order - hence getDeclaredMethods
+        for (Method method : beanType.getDeclaredMethods()) {
+            // note: the return type void requirement is relaxed (allow chained setters)
+            if (setterName.equals(method.getName())
+                    && Modifier.isPublic(method.getModifiers())
+                    && !method.isBridge()
+                    && method.getParameterCount() == 1
+                    && method.getParameterTypes()[0] == propertyType) {
+                return method;
+            }
+        }
+        Class<?> superClass = beanType.getSuperclass();
+        if (superClass == null || superClass == Object.class) {
+            return null;
+        } else {
+            return getSetter(superClass, propertyName, propertyType);
         }
     }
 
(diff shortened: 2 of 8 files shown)


easy-random-core/pom.xml
```diff
@@ -57,10 +57,6 @@
             <groupId>io.github.classgraph</groupId>
             <artifactId>classgraph</artifactId>
         </dependency>
-        <dependency>
-            <groupId>commons-beanutils</groupId>
```
seregamorph, line 61: there is one more problem with `commons-beanutils`: it requires to use the latest version and enforces version upgrade. This project respects back compatibility, but I still had compilation issues related to generic types. In other words, if this dependency is kept, I strongly recommend to shade it inside of easy-random jar instead of dependency

easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
```diff
@@ -140,21 +131,16 @@ public Object invoke(final Object proxy, final Method method, final Object[] arg
...
+        try {
+            Method setter = getSetter(object.getClass(), field.getName(), field.getType());
             if (setter != null) {
                 setter.invoke(object, value);
             } else {
                 setFieldValue(object, field, value);
             }
-        } catch (IntrospectionException | IllegalAccessException e) {
-            try { // to call fluent setter
-                PROPERTY_UTILS_BEAN.setProperty(object, field.getName(), value);
```
seregamorph, line 153: I've tried the suggested option to flush caches of `java.beans.Introspector` and recreated the `PropertyUtilsBean` object, but it allows only to address the first issue (with overriden setter), but not the second (with overriden field)

easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
```diff
@@ -140,21 +131,16 @@ public Object invoke(final Object proxy, final Method method, final Object[] arg
...
-            } catch (NoSuchMethodException noSuchMethodException) {
-                // otherwise, set field using reflection
-                setFieldValue(object, field, value);
-            }
+        } catch (IllegalAccessException e) {
```
seregamorph, line 141: probably this will never happen (public method filter is used)

easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
```diff
@@ -610,6 +596,27 @@ private static void rejectUnsupportedTypes(Class<?> type) {
         }
     }
 
+    private static Method getSetter(Class<?> beanType, String propertyName, Class<?> propertyType) {
+        String setterName = "set" + propertyName.substring(0, 1).toUpperCase(ENGLISH) + propertyName.substring(1);
+        // implementation note: we go from subclass to super-class order - hence getDeclaredMethods
+        for (Method method : beanType.getDeclaredMethods()) {
```
seregamorph, line 602: Following KISS principle, this may be simplified to `beanType.getMethods()` instead (in this case the recursive super call can be omitted).

I've validated this change on about 200 model classes in my projects (I had good coverage with cases - so I've found issues described in #437), so at least I'm sure that it's better than it was in `4.2.0` and in `master`

easy-random-core/src/test/java/org/jeasy/random/beans/ChainedSetterBean.java
```diff
@@ -23,16 +23,27 @@
  */
 package org.jeasy.random.beans;
 
-public class FluentSetterBean {
+public class ChainedSetterBean {
```
seregamorph, line 26: The reason why this class is renamed.

"Fluent" word was used because of `commons-beanutils` `FluentPropertyBeanIntrospector`. Now it's not related.

The lombok project suggests these definitions:
**Fluent accessors**:
```java
class Data {

    private String value;

    public String value() {
        return value;
    }

    public Data value(String value) {
        this.value = value;
        return value;
    }
}
```

**Chained accessors**:
```java
class Data {

    private String value;

    public String getValue() {
        return value;
    }

    public Data setValue(String value) {
        this.value = value;
        return value;
    }
}
```

easy-random-core/src/test/java/org/jeasy/random/beans/ChainedSetterBean.java
```diff
@@ -23,16 +23,27 @@
...
 
-    public String name;
```
seregamorph, line 28: this obviously should not be `public`

easy-random-core/src/test/java/org/jeasy/random/beans/ChainedSetterBean.java
```diff
@@ -23,16 +23,27 @@
...
+    private String name;
+    private int index;
```
seregamorph, line 29: we cover several special cases: primitive type, single character field name

easy-random-core/pom.xml
seregamorph: Also it can cause silent problem (no compilation issue) with runtime failure in production (if the code that calls easy-random is not checked in tests).

easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
seregamorph: Decided to simplify (see second commit) and unify with `getReadMethod`

easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
```diff
@@ -569,9 +565,13 @@ private static void rejectUnsupportedTypes(Class<?> type) {
         return getPublicMethod("is" + capitalizedFieldName, fieldClass);
     }
 
-    private static Optional<Method> getPublicMethod(String name, Class<?> target) {
+    private static String capitalize(String propertyName) {
```
seregamorph, line 568: no need to validate string length (private usage for field names only)

benas:
This is awesome!  Rebased and merged as e86984d6052e80913745918467a4d0ec1155f044. Thank you for all the effort you've put into this 👍

easy-random-core/src/main/java/org/jeasy/random/util/ReflectionUtils.java
```diff
@@ -551,6 +537,16 @@ private static void rejectUnsupportedTypes(Class<?> type) {
         }
     }
 
+    /**
+     * Get the write method for given field.
+     *
+     * @param field field to get the write method for
+     * @return Optional of write method or empty if field has no write method
+     */
+    private static Optional<Method> getWriteMethod(Field field) {
```
benas, line 546: This could public to be consistent with `getReadMethod` which is public.  I already fixed that on merge.

PR #439, opened 2020-11-05
Add support for generic type variables

benas:
This PR is for issue #425.

It covers basic use cases of randomizing generic fields where the type is defined at the class level (or one of its base types), like:

```java
@Test
void testGenericTypeRandomization() {
    // given
    class Base<T> {
        T t;
    }
    class Concrete extends Base<String> {}
    
    // when
    Concrete concrete = easyRandom.nextObject(Concrete.class);
    
    // then
    assertThat(concrete.t).isInstanceOf(String.class);
    assertThat(concrete.t).isNotEmpty();
}
```

It also supports multiple generic types and multiple generic levels in the hierarchy (see tests). However, it does not covers complex types such as `class Concrete extends Base<List<String>> {}` in the previous example. This will be documented accordingly.

As discussed in #426 , this is a good start and may cover most use cases. Custom randomizers can be used for specific cases. Any feedback is welcome.

### DIFF
--- a/easy-random-core/src/main/java/org/jeasy/random/FieldPopulator.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/FieldPopulator.java
@@ -23,7 +23,10 @@
  */
 package org.jeasy.random;
 
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.TypeVariable;
 import java.util.List;
+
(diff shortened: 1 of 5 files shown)


benas:
rebased and merged as beab83ce4fd2e3b2730826fe0374bca603ae64ee.

Issue #447, opened 2020-12-03
Enhance documentation of Bean Validation constraint parameters precedence


soulsreaver666:
In case a get method for some List in the objects is annotated with : 
@Size(min=1)   public List<Address> getAddresses() {
    return addresses;
  }

Easy random do not take into consideration collection size limitation I provide : 
EasyRandomParameters parameters = new EasyRandomParameters()
            .seed(random.nextLong())
            .randomizationDepth(2)
            .collectionSizeRange(1, 4)
            .overrideDefaultInitialization(true);

 BusinessPartner businessPartner; = new EasyRandom(parameters).nextObject(BusinessPartner.class);

and generates really long list

![image](https://user-images.githubusercontent.com/74959109/100985732-9778e000-354c-11eb-82bd-b17398eaae0e.png)

Is there a way to limitate such kind of list ?

murdos:
@soulsreaver666 : I don't think there's a way with current released versions.

I'll soon push a PR that allows that, by attempting to reduce range from `@Size` annotation, using collectionSizeRange from EasyRandomParameters.

soulsreaver666:
Great. Just let me know

benas:
> In case a get method for some List in the objects is annotated with @SiZe(min=1) .. Easy random do not take into consideration collection size limitation I provide

This is the expected behaviour which is documented (but not clearly) in the [bean validation](https://github.com/j-easy/easy-random/wiki/bean-validation-support) section of the wiki, see:

```
Heads up: Bean validation constraints will take precedence over ...
```

I will update the wiki to make this more explicit. So Bean Validation annotation parameters take precedence over global parameters, otherwise Easy Random will generate **invalid** data, which would obviously be incorrect. Here is a test case that shows the expected behaviour:

```java
@Test
void beanValidationParametersShouldTakePrecedenceOverGlobalParameters() {
    // given
    class Person {
        @Size(min = 15, max = 20)
        private List<String> names;
    }
    EasyRandomParameters parameters = new EasyRandomParameters().collectionSizeRange(1, 10);
    EasyRandom easyRandom = new EasyRandom(parameters);
    
    // when
    Person person = easyRandom.nextObject(Person.class);
    
    // then
    assertThat(person.names.size()).isBetween(15, 20);
}
```

In your case, the range from the `@Size` annotation is [1, 255] (the max defaults to 255), that's why ER generates a value (241) in that range and not in the [1, 4] range from the `collectionSizeRange` parameter.

[...]

PR #448, opened 2020-12-14
@Size constraints should not be propagated to generated collection items

murdos:
The BeanValidation constraints of `@Size` are incorrectly used to configure the new EasyRandom instance used by `SizeAnnotationHandler` for both collection size and string length range , with the effects of applying wrong range to generated fields that are embedded.

The following test illustrates the issue:
```java
    @Test
    void test() {
        
        class ChildBean {
            public String string;
        }
        class ParentBean {
            @Size(min = 2, max = 5)
            public List<ChildBean> childs;
        }

        EasyRandomParameters parameters = new EasyRandomParameters().stringLengthRange(10, 15);
        EasyRandom random = new EasyRandom(parameters);
        ParentBean parentBean = random.nextObject(ParentBean.class);

        assertThat(parentBean.childs).allSatisfy(child -> assertThat(child.string).hasSizeBetween(10, 15));
    }
```

I stumbled upon this bug while investigating issue #447

### DIFF
--- a/easy-random-bean-validation/src/main/java/org/jeasy/random/validation/SizeAnnotationHandler.java
+++ b/easy-random-bean-validation/src/main/java/org/jeasy/random/validation/SizeAnnotationHandler.java
@@ -62,8 +62,6 @@ class SizeAnnotationHandler implements BeanValidationAnnotationHandler {
         final int min = sizeAnnotation.min();
         final int max = sizeAnnotation.max() == Integer.MAX_VALUE ? 255 : sizeAnnotation.max();
         if (easyRandom == null) {
-            parameters.setCollectionSizeRange(new EasyRandomParameters.Range<>(min, max));
-            parameters.setStringLengthRange(new EasyRandomParameters.Range<>(min, max));
             easyRandom = new EasyRandom(parameters);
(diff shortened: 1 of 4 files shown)


benas:
Thank you for your PR! The description does not explain the problem and the title is too short to explain it clearly. Since you did not create a separate issue, this PR will eventually be used in the change log, and people won't be able to understand the problem from the title and shouldn't be looking at the code changes of the PR to understand the issue. Could you please edit the description and clarify the issue this PR is solving? Thank you upfront.

murdos:
@benas : description amended, I hope it's good enough.

benas:
Thank you for these updates! LGTM, rebased and merged as 0e56afa97f89b58c4df9edd8eee8f900efa3a4ba. There is another case we need to cover which is when BV constraints on a bean should take precedence over configuration parameters (as documented in the [wiki](https://github.com/j-easy/easy-random/wiki/bean-validation-support#bean-validation-contraint-parameters-precedence)). I added a test for that (see 0426dcd8b2df555d276a16513ab92d79f99f8de5) which passes with this PR as well 👍 

PR #450, opened 2020-12-16
Add regression test for generic fields

reitzmichnicht:
This adds a test to spot a regression in release 5.0.0

The wrong context is used to resolve the generic parameters. Had no success yet to also find a fix.


benas:
@reitzmichnicht Thank you for reporting this issue! I created #452 to track it as I can't merge this PR for now since it breaks the build (I prefer an open issue over an open incomplete PR). Please re-open this PR or create a new one if you manage to find a fix.

Issue #449, opened 2020-12-15
Possibility to disable cache


mjureczko:
EasyRandom caches the generated objects to speed things up when numerous objects are requested. However, it is not a perfect match for every use case. Sometimes one may need more unique objects. Reusing the same objects may create inter tests dependencies when using easy-random for generating test data, which can be especially nasty when the objects are mutable. When using recursive data structures, the cache can create infinite loops (A->B->C->B...)

Currently, the only possibility to not use cache is to set `objectPoolSize` to a large value. Effectively, it can cause that the cache will never fill up and as consequence, will never be used to fetch an object. However, such an approach has a significant and unnecessary memory footprint - each object created by the easy-random will be kept in memory, but none will be reused.

If an additional property for disabling cache is in line with the vision of this project, I will be happy to provide an MR with such solution.

JohannesBe:
For my back-reference implementation (see https://github.com/j-easy/easy-random/issues/444) I would also like the possibility of turning the cache off. Might I suggest using a value of `0` for the `objectPoolSize` in order to accomplish this? I feel this would have non-breaking impact* whilst keeping the API-compatibility maximal at the same time.

\* this value is currently completely impossible, since the setter has a check for this:
```
    public void setObjectPoolSize(int objectPoolSize) {
        if (objectPoolSize < 1) {
            throw new IllegalArgumentException("objectPoolSize must be >= 1");
        }
        this.objectPoolSize = objectPoolSize;
    }
```

JohannesBe:
There might be more strings attached to doing this however, see the docstring of `RandomizationContext`:
>  * It contains a map acting as a cache of populated beans to avoid infinite recursion.

Not sure what that's about.

ldmi3i:
> There might be more strings attached to doing this however, see the docstring of `RandomizationContext`:
> 
> > * It contains a map acting as a cache of populated beans to avoid infinite recursion.
> 
> Not sure what that's about.

I think if there is class A and class B. And class A have field B b , class B have field A a . It causes infinite recursion because without using of pool each class will try populate this reference with new object instead of using already populated. Pool bounds process of creation new beans with pool size.
So disabling of pool is dangerous and can cause infinite recursion.

JohannesBe:
But isn't that what the recursion depth setting is for?

mjureczko:
[...]

PR #451, opened 2020-12-20
Add option to disable object pool cache #449

JohannesBe:
```
[INFO] Reactor Summary for Easy Random 5.0.1-SNAPSHOT:
[INFO] 
[INFO] Easy Random ........................................ SUCCESS [  0.590 s]
[INFO] Easy Random Core ................................... SUCCESS [ 25.148 s]
[INFO] Easy Random Randomizers ............................ SUCCESS [  3.562 s]
[INFO] Easy Random Bean Validation ........................ SUCCESS [  4.638 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  34.109 s
[INFO] Finished at: 2020-12-20T15:16:01+01:00
[INFO] ------------------------------------------------------------------------
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,7 @@ If you believe you found a bug or have any question, please use the [issue track
 * [Unconditional One](https://github.com/unconditional)
 * [JJ1216](https://github.com/JJ1216)
 * [Sergey Chernov](https://github.com/seregamorph)
+* [Johannes Bergé](https://github.com/JohannesBe)
 
 Thank you all for your contributions!
 
--- a/easy-random-core/src/main/java/org/jeasy/random/EasyRandomParameters.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/EasyRandomParameters.java
@@ -177,8 +177,8 @@ public class EasyRandomParameters {
         return objectPoolSize;
     }
     public void setObjectPoolSize(int objectPoolSize) {
-        if (objectPoolSize < 1) {
-            throw new IllegalArgumentException("objectPoolSize must be >= 1");
+        if (objectPoolSize < 0) {
+            throw new IllegalArgumentException("objectPoolSize must be >= 0");
         }
         this.objectPoolSize = objectPoolSize;
     }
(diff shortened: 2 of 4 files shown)


JohannesBe:
As mentioned in #449 I do not exactly know what the implications of this MR could be for recursive objects (StackOverflows?). Would be great if you could shine a light on that @benas :)

ldmi3i:
I think you can add test like this to test if recursion happens or not.

```
    @Test
    void recursionTest() {
        EasyRandomParameters par = new EasyRandomParameters();
        par.objectPoolSize(0);
        EasyRandom easyRandom = new EasyRandom(par);
        easyRandom.nextObject(A.class);
    }

    private static class A {
        private B b;

        public B getB() {
            return b;
        }

        public void setB(B b) {
            this.b = b;
        }
    }

    private static class B {
        private A a;

        public A getA() {
            return a;
        }

        public void setA(A a) {
            this.a = a;
        }
    }
```

mjureczko:
I believe that it solves the issue correctly, but it will do no harm to add the test case suggested by @ldmi3i

JohannesBe:
I will add the test later tonight, or later this week.

JohannesBe:
I have added this test. This test in fact did result in a `StackOverFlowException` if the randomizationDepth was large (e.g. initial value `Integer.MAX_VALUE`). Therefore I have added a check that verifies the randomizationDepth is set when disabling the cache.

If this PR is merged, I propose we also edit the [wiki](https://github.com/j-easy/easy-random/wiki/Randomization-parameters) to document the changes in the configuration.

JohannesBe:
I am not entirely happy with this inter-dependency of configuration parameters however, users might still set their configuration in a broken state and the consequence might not be apparent to the user at all. Any recommendations?

mjureczko:
I'm not convinced that the large randomizationDepth issue is in fact issue. I do not expect anything but `StackOverFlowException` when using `Integer.MAX_VALUE`. Besides the relation between cache size (objectPoolSize) and randomizationDepth was always there as the cache usages do not respect randomizationDepth.

benas:
@JohannesBe I really appreciate the effort you've put into this, but as explained in #449 , I see no need to change the code since there is already a way to disable caching. I always recommend discussing with project maintainers about opening a PR for a feature request before actually starting the work.

Thanks again for your time and effort.

Issue #461, opened 2021-07-06
Initialized empty Collection is not populated


BenVercammen:
In our project, we work with generated classes that contain List properties. 

At first this code was generated from WSDL's, and the List properties would not be initialized in the generated code. 
``protected List<SomeType> list;``

However, we are now generating the code from an OpenAPI spec and the List properties are already initialized.
``private List<SomeType> list = new ArrayList<>();``

This causes an issue as these lists are not populated because they are not null anymore.

I've already created a test and a fix, I'll open a pull request in a minute.

fmbenhassine:
Thank you for opening this issue. However, as mentioned by @msievers , you can use the `overrideDefaultInitialization`, which is designed specifically for that case. Here is the section in the docs that explains that parameter with an example: https://github.com/j-easy/easy-random/wiki/Randomization-parameters#override-default-initialization.

I am closing this issue and the corresponding PRs as I believe the solution is already implemented with the `overrideDefaultInitialization` parameter. Thank you for your time anyway.

@msievers Thank you for the follow up!

PR #462, opened 2021-07-06
#461 populate already initialized Collections

BenVercammen:
Proposed fix for #461 

Please note that this branch is not yet rebased on the latest version of the master branch, as I'm not sure how the 4.3.x vs 5.0.x branches are supposed to be set up. We're using the 4.3.x version at the moment, so I've branched from the tagged commit.
Any pointers on how to properly fix this are welcome!

### DIFF
--- a/easy-random-bean-validation/pom.xml
+++ b/easy-random-bean-validation/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <artifactId>easy-random</artifactId>
         <groupId>org.jeasy</groupId>
-        <version>4.3.0</version>
+        <version>4.3.1-SNAPSHOT</version>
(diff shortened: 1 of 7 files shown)


kobaevanda20:
#489

fmbenhassine:
Thank you for your PR! However, as explained in #461, the way to address this case is by using the `overrideDefaultInitialization` parameter. So I am closing this PR. Appreciated your time anyway.

As a side note, v4.3 and v5.0 won't be maintained. I am currently working on v6 and unfortunately I don't have time to maintain several versions. `overrideDefaultInitialization` is present in 4.3 and 5.0 anyway, so you can use it if you are still on those versions.

PR #463, opened 2021-07-07
#461 populate already initialized Collections

BenVercammen:
Proposed fix for #461

This branch is rebased on the current master branch. However, we need the 4.3.x version, hence also pull request #462 ...


fmbenhassine:
Thank you for your PR! However, as explained in #461, the way to address this case is by using the `overrideDefaultInitialization` parameter. So I am closing this PR. Appreciated your time anyway.

As a side note, v4.3 and v5.0 won't be maintained. I am currently working on v6 and unfortunately I don't have time to maintain several versions. `overrideDefaultInitialization` is present in 4.3 and 5.0 anyway, so you can use it if you are still on those versions.

PR #468, opened 2021-11-07
Log skip field population warning

astubbs:
Just burnt quite a bit of time trying to grok the lib and find out why my objects were ending up with null fields. Turns out my recursive structure's depth limit wasn't high enough do satisfy my desired collection size.

Once I'd understood that classes can get instantiated with null fields (https://github.com/j-easy/easy-random/issues/467), I found where field setting was being skipped for some objects in my collections.

I think adding a log warning for these points would be very helpful, especially for new users.

I'm a bit surprised to see there's no logging anywhere. Are you opposed to adding any?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Java CI
 
-on: [push]
+on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


fmbenhassine:
I am not opposed to add logging. And it makes sense to me to add a warning for skipped fields. Thank you for raising this.

The PR LGTM. I just replaced `logback` with `slf4j-simple` for tests. Rebased and merged as 51150d46. Thank you for your contribution!

PR #473, opened 2022-03-02
#472 Allow using EnumRandomizer with excluded values deterministically

### LINKED ISSUE #472: Cannot use EnumRandomizer with excluded values deterministically
The `EnumRandomizer` has a constructor which lets you specify an `enum` class and a set of values to exclude. This works great! However, it's not deterministic when used this way, as there's no way to set the seed for its internal `Random` instance. 

What would be great is adding the `seed` argument to the constructor, e.g.

```java
    /**
     * Create a new {@link EnumRandomizer}.
     *
     * @param enumeration    the enumeration from which this randomizer will generate random values
     * @param seed                the initial seed
     * @param excludedValues the values to exclude from random picking
     * @throws IllegalArgumentException when excludedValues contains all enumeration values,
     *                                  ie all elements from the enumeration are excluded
     */
    public EnumRandomizer(final Class<E> enumeration, final long seed, final E... excludedValues) throws IllegalArgumentException {
        super(seed);
        checkExcludedValues(enumeration, excludedValues);
        this.enumConstants = getFilteredList(enumeration, excludedValues);
    }
```

This way, we can reuse this in a deterministic way, such that randomized input can be regenerated on demand for debugging purposes.

npepinpe:
This PR allows using `EnumRandomizer` with excluded values in a deterministic fashion, by also passing in a seed. This was previously only possible when using all the values from the given `enum`.

closes #472

### DIFF
--- a/easy-random-core/src/main/java/org/jeasy/random/randomizers/misc/EnumRandomizer.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/randomizers/misc/EnumRandomizer.java
@@ -76,6 +76,21 @@ public class EnumRandomizer<E extends Enum<E>> extends AbstractRandomizer<E> {
     }
 
     /**
+     * Create a new {@link EnumRandomizer}.
+     *
+     * @param enumeration    the enumeration from which this randomizer will generate random values
+     * @param seed           the initial seed
+     * @param excludedValues the values to exclude from random picking
+     * @throws IllegalArgumentException when excludedValues contains all enumeration values,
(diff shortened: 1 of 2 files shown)


murdos (approved):

armandino:
@npepinpe also a great idea! I will implement this in [Instancio](https://github.com/instancio/instancio/) as well.

fmbenhassine:
LGTM 👍  Thank you for raising the issue and contributing a PR!

PR #474, opened 2022-03-02
#471 Allow flexible concrete type resolution

### LINKED ISSUE #471: Restrict scanning for concrete types over specific packages
Class scanning is a great feature, but depending on the size of your project, it leads to some minor issues. For example, I'd like to use `EasyRandom` as part of the test harness for a project. The test harness has its own module as it gets included into different module. Scanning for types may end up finding different implementations of the interfaces depending on the module in which the test harness is pulled. I'm sure that's a nice use case for some, but it can also lead to some issues. For example, the test harness module already pulls in a dependency which has concrete implementations of some types. These are beans and can easily be generated by `EasyRandom`. However, maybe in another module, there are other implementations which run into limitations of `EasyRandom`, or which aren't strictly beans. In this case, it would be great if I could specify which implementations to use.

Secondly, scanning the whole class path can be prohibitively expensive. This isn't a big overhead when running a complete test suite, but when running single tests in your IDE having it add a few seconds for a test which takes 5 milliseconds is a bit tedious.

Now, neither are blockers, but restricting the scanning to specific packages seems like a pretty easy fix - this is already supported by `ClassGraph`, and it would fix both issues.

Let me know if I wasn't clear. If this is something you'd consider, I'm also happy opening the PR myself.

npepinpe:
This PR introduces a new `TypeResolver` interface to the public API. This interface's default implementation is in `ReflectionUtils`, which is then delegated to the static helpers in `ClassGraphFacade`. This means the new feature remains backwards compatible with previous behavior.

Users can now specify a new parameter, `EasyRandomParameters#typeResolver(TypeResolver)`, which lets them specify how concrete types will be resolved without coupling the library to `ClassGraph` in any way, though users can still use `ClassGraph` to implement it themselves.

An internal proxy class is used to wrap custom instances of `TypeResolver` set via the parameters. This is mostly because when creating the default `ObjenesisObjectFactory`, we need to pass it the current type resolver. However, it may get modified later on. So instead, we pass a proxy, which we can always modify thereafter. The proxy is not accessible to the user, so the only way to modify it is via the parameters. The downside here is that if a user modifies the parameters post-creation of `EasyRandom`, it will affect the internal `ObjenesisObjectFactory` and change the type resolver. I think that's a minor downside, as I don't expect this to be the common case, and I would hope users make use of the `copy()` factory method when reusing parameters. The `copy()` method will not copy the proxy but the actual `TypeResolver`, so it will really be a copy.

closes #471

### DIFF
--- a/easy-random-core/src/main/java/org/jeasy/random/EasyRandom.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/EasyRandom.java
@@ -83,7 +83,8 @@ public class EasyRandom extends Random {
         enumRandomizersByType = new ConcurrentHashMap<>();
         fieldPopulator = new FieldPopulator(this,
                 this.randomizerProvider, arrayPopulator,
-                collectionPopulator, mapPopulator, optionalPopulator);
+                collectionPopulator, mapPopulator, optionalPopulator,
+                easyRandomParameters.getTypeResolver());
         exclusionPolicy = easyRandomParameters.getExclusionPolicy();
         parameters = easyRandomParameters;
     }
--- a/easy-random-core/src/main/java/org/jeasy/random/EasyRandomParameters.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/EasyRandomParameters.java
@@ -33,6 +33,7 @@ import java.nio.charset.StandardCharsets;
 import java.time.*;
 import java.util.*;
 import java.util.function.Predicate;
+import org.jeasy.random.util.TypeResolverProxy;
 
 import static java.lang.String.format;
 import static java.time.ZonedDateTime.of;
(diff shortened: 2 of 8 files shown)


armandino:
@npepinpe that's an interesting suggestion. I will try to implement your idea in Instancio. It's an alternative library that offers similar functionality to Easy Random. Since Easy Random is no longer maintained, I will link it here so others can find it.

https://github.com/instancio/instancio

fmbenhassine:
Thank you for your PR! I agree on the introduction of `TypeResolver`. However I am not a big fan a wrapping the default one in a proxy, I would keep it simple and just set the default where appropriate and make it configurable.

Please rebase the PR on the latest `master` and it should be good to merge. Thank you upfront.

fmbenhassine:
As mentioned in #471, I am not planning to include this in Easy Random with the project being in maintenance mode. I shared my attempt to this here: https://github.com/j-easy/easy-random/issues/471#issuecomment-4606982763.

Please choose whatever solution that works best for you and add it in your fork.

Thank you for the PR anyway.

PR #482, opened 2022-09-05
Expose actual randomizer provider in RandomizerContext

murdos:
This will allow custom (ContextAware)Randomizers or custom field populators to access registered randomizers, preventing them the pain of reimplementing registries loading.

### DIFF
--- a/easy-random-core/src/main/java/org/jeasy/random/EasyRandom.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/EasyRandom.java
@@ -101,7 +101,7 @@ public class EasyRandom extends Random {
         if (type.isRecord()) {
             return createRandomRecord(type);
         } else {
-            return doPopulateBean(type, new RandomizationContext(type, parameters));
+            return doPopulateBean(type, new RandomizationContext(type, parameters, randomizerProvider));
         }
     }
 
--- a/easy-random-core/src/main/java/org/jeasy/random/RandomizationContext.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/RandomizationContext.java
@@ -24,6 +24,7 @@
 package org.jeasy.random;
 
 import org.jeasy.random.api.RandomizerContext;
+import org.jeasy.random.api.RandomizerProvider;
 
 import java.lang.reflect.Field;
 import java.util.*;
@@ -38,6 +39,8 @@ import static java.util.stream.Collectors.toList;
  */
 class RandomizationContext implements RandomizerContext {
 
+    private final RandomizerProvider randomizerProvider;
+
     private final EasyRandomParameters parameters;
 
     private final Map<Class<?>, List<Object>> populatedBeans;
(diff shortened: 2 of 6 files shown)


murdos:
@fmbenhassine : is there any chance you could include this in 6.0 ?

fmbenhassine:
There is always a chance! 😄 

However, it is already possible to get the `RandomizerProvider` through the parameters (`EasyRandomParameters#getRandomizerProvider`), which are present in the context. Do you agree?

easy-random-core/src/main/java/org/jeasy/random/EasyRandom.java
```diff
@@ -97,7 +97,7 @@ public EasyRandom(final EasyRandomParameters easyRandomParameters) {
      * @throws ObjectCreationException when unable to create a new instance of the given type
      */
     public <T> T nextObject(final Class<T> type) {
-        return doPopulateBean(type, new RandomizationContext(type, parameters));
+        return doPopulateBean(type, new RandomizationContext(type, parameters, randomizerProvider));
```
fmbenhassine, line 100: No need to pass the `randomizerProvider` as a third parameter here, since it can be accessed through the parameters, which are part of the context.

murdos:
> There is always a chance! 😄
> 
> However, it is already possible to get the `RandomizerProvider` through the parameters (`EasyRandomParameters#getRandomizerProvider`), which are present in the context. Do you agree?

`EasyRandomParameters#getRandomizerProvider` can only be used to get the custom `RandomizerProvider` that may have been configured.
My intent through this feature request is to be able to access to the actual `RandomizerProvider` that is used by `EasyRandom` : 
https://github.com/j-easy/easy-random/blob/4b46b05a44cf1f606624a34fb783b31816107344/easy-random-core/src/main/java/org/jeasy/random/EasyRandom.java#L77

easy-random-core/src/main/java/org/jeasy/random/EasyRandom.java
murdos: `randomizerProvider` and `parameters.getRandomizerProvider()` are different, and in canonical use, `parameters.getRandomizerProvider()` will be `null`:
https://github.com/j-easy/easy-random/blob/4b46b05a44cf1f606624a34fb783b31816107344/easy-random-core/src/main/java/org/jeasy/random/EasyRandom.java#L77

fmbenhassine: Well this is the problem then, they should _not_ be different. You should get either the custom provider if any, or the default one, but not null. This is actually an initialization issue in ER, which was due to a wrong assertion (1912e107f1d2cce4313d3993b552f72308f970e1) and which I fixed in f0b8851f28ed88ed7131d694034a5d521d374be5. Can you please give the snapshots a try? You should now be able to get the randomizer provider from the parameters.

murdos: Great, this will indeed help solving my problem!
I can now close this PR :)

Issue #483, opened 2022-11-16
[Bug] dateRange failed to handle LocalDate.MAX


lukeZhangMengxi:
To reproduce, simply create the parameter as 
```
import java.time.LocalDate
import org.jeasy.random.EasyRandomParameters

EasyRandomParameters().dateRange(LocalDate.now(), LocalDate.MAX)
```
which will fail with error
```
max must be after min
java.lang.IllegalArgumentException: max must be after min
	at org.jeasy.random.randomizers.range.DateRangeRandomizer.checkValues(DateRangeRandomizer.java:59)
	at org.jeasy.random.randomizers.range.AbstractRangeRandomizer.<init>(AbstractRangeRandomizer.java:50)
	at org.jeasy.random.randomizers.range.DateRangeRandomizer.<init>(DateRangeRandomizer.java:53)
	at org.jeasy.random.randomizers.registry.InternalRandomizerRegistry.init(InternalRandomizerRegistry.java:98)
	at org.jeasy.random.EasyRandom.lambda$setupRandomizerRegistries$2(EasyRandom.java:218)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.jeasy.random.EasyRandom.setupRandomizerRegistries(EasyRandom.java:218)
	at org.jeasy.random.EasyRandom.<init>(EasyRandom.java:74)
```

I suspect there is some integer overflow issue.

fmbenhassine:
Thank you for reporting this! Fixed in #484 .

PR #484, opened 2022-11-27
Fix for LocalDate.MIN and LocalDate.MAX range value exceptions #483

OguzhanKazak:
Proposed fix for #483


lukeZhangMengxi (approved):
looks good to me

fmbenhassine:
LGTM. Thank you for your PR!

PR #491, opened 2023-03-14
Make usage of ClassGraphFacade replacable with custom logic in ReflectionUtils

pokab:
...by providing user injectable Function, while retaining ClassGraphFacade's lazy loading semantics.

Moves concrete subtype cache to ReflectionUtils.

Additionally:
- All tests pass (as I have built a snapshot for myself).
- I didn't create an issue. Should I?

### DIFF
--- a/easy-random-core/src/main/java/org/jeasy/random/util/ClassGraphFacade.java
+++ b/easy-random-core/src/main/java/org/jeasy/random/util/ClassGraphFacade.java
@@ -25,7 +25,6 @@ package org.jeasy.random.util;
 
 import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.ConcurrentHashMap;
 
 import io.github.classgraph.ClassGraph;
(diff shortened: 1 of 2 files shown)


fmbenhassine:
Thank you for your PR. However, why should we make classgraph replaceable? If I follow this reasoning, we should also find a way to make objenesis replaceable, and datafker replaceable as well.

If you want to customize the way to scan the classpath for concrete types, you can disable the `scanClasspathForConcreteTypes` parameter and provide a custom object factory. I see no need to add an indirection to make a dependency swappable. Do you agree?

fmbenhassine:
This will be resolved with #474.

Thank you for your contribution anyway.

pokab:
My reason behind this change had to do with a requirement in our test stack, where we had to filter classes on the classpath. The solution in #474 (which is not yet merged) should provide the same possibility. Thank you.

PR #493, opened 2023-05-05
[BUGFIX] Make BeanValidationAnnotationHandler public

shalak:
That way, extending BeanValidationRandomizerRegistry and appending to annotationHandlers is possible.

This is a follow up for change mentioned [here](https://github.com/j-easy/easy-random/issues/401#issuecomment-604739142) - making the map protected was unfortunately, not enought. The [test was passing](https://github.com/j-easy/easy-random/commit/159869580487e2ecf16bcc1fd890564b0c4d3fcd#diff-b397b0a3111c419b2b61e7e667a04d9c02bc9ecaa091a674fac904250b6211bdR35) because `BeanValidationAnnotationHandler` is in the same package as the test.


fmbenhassine:
Thank you for the PR, LGTM. Even though the bean validation module will be discontinued (see #499), I am going to merge this change so that it is included in the module before removing it (for someone else to maintain or copy).

Issue #499, opened 2023-07-15
Remove Bean Validation module


fmbenhassine:
This module has been a maintenance burden over the years and during each major release. Therefore, it will not be released starting from v6. 

If someone is interested in maintaining it as an extension, please show interest by adding a comment here and I will mention it in the [extensions section](https://github.com/j-easy/easy-random#extensions) of the README.

Otherwise, the code of the module can be copied in projects that are interested in using it. After all, that module can be used as an extension by just adding the `BeanValidationRandomizerRegistry` (+ the required annotation handlers) to the classpath and it will be automatically discovered and used by ER.

pantherwelt:
Hi, sorry, first time here. 
We have been using easy-random 5 with bean validation.
Since we upgraded our project now we need it to work with the "jakarta" annotations.
I read your instructions in the comment above.
So, where is actually the code for the bean validation module ?

PR #505, opened 2023-08-18
Make validation handlers public

shalak:
* This will allow users to create a BeanValidationRandomizerRegistry implementation with a subset of handlers


shalak:
This further improves developer experience along with https://github.com/j-easy/easy-random/pull/493

Is there a plan to release current `master` build?

fmbenhassine:
Bean validation support will be removed in v6, #499. Therefore, this can be done in a fork. Thank you for your contribution anyway!

PR #512, opened 2023-12-02
dependencies version up; jitpack.io

magicprinc:
- Fix jitpack.io
- dependencies version up

### DIFF
--- a/jitpack.yml
+++ b/jitpack.yml
@@ -1,0 +1,4 @@
+jdk:
+  - openjdk17
+install:
+  - mvn install -Dmaven.javadoc.skip=true -DskipTests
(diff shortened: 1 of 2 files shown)


magicprinc:
https://jitpack.io/ must work with maven projects, but it didn't.
With `jitpack.yml` from this PR jitpack.io works as expected. 
See: https://jitpack.io/#magicprinc/easy-random/5.2

Many people still can't upgrade from Slf4j 1.7.+ to 2.+
If you compile with Slf4j 1.7.+ it works with 2.+

I also upgraded all other dependencies to current versions

Tests fails on JDK21, but there is a problem with Mockito and JDK21

codeCraft-Ritik (commented):
Great work! The implementation is clean and easy to understand

fmbenhassine:
Please add that `jitpack.yml` file in your fork, there is no need to include it here.

All dependencies were updated to the latest versions in v6. If you need earlier versions, please do it in your fork and serve the jar from jitpack.

Thank you for the PR anyway.

PR #513, opened 2024-02-29
Add advanced generic and wild-card parse mechanism

amakaroff:
Hello!
We use your library for generating random values for our model. Our model has difficult generic and easy-random can't determinate some generic and skipped it. But we require fill all models by non null or empty values. So we decided to fix it in easy-random
From our side, it's look like a fix, but changes a little large :)
We understand about maintenance mode in easy-random, but I hope our changes will accepted like fix. In otherwise we will have to fork it. But we don't want to create our fork. I add that changes by flag, so current behavior not changed.

Thank you for library!

---

Thank you for taking time to contribute this pull request :smile:

You have probably already checked the following points, but as a reminder, please ensure that:

* your branch is rebased on the latest version of the master branch
* all tests pass
* code is formatted with IntelliJ Idea default settings (if you use another IDE, never mind, ignore this point)

You may add the reference of the issue in your commit message.

Don't hesitate to add your name to the contributors list in the `README.md` file.

Many thanks upfront!

### DIFF
--- a/src/main/java/org/jeasy/random/CollectionPopulator.java
+++ b/src/main/java/org/jeasy/random/CollectionPopulator.java
@@ -41,15 +41,18 @@ class CollectionPopulator {
 
     private final EasyRandom easyRandom;
 
-    CollectionPopulator(final EasyRandom easyRandom) {
+    private final GenericResolver genericResolver;
+
+    CollectionPopulator(final EasyRandom easyRandom, final GenericResolver genericResolver) {
         this.easyRandom = easyRandom;
+        this.genericResolver = genericResolver;
     }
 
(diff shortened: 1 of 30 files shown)


prenevin:
Excited for this merge

codeCraft-Ritik (commented):
Great work! The implementation is clean and easy to understand

fmbenhassine:
I think the current generics support in Easy Random already covers most cases, but if you need more advanced support, then you should be able to do that by providing a smarter `ObjectFactory`. If that is still not enough, then you can apply this PR to a fork. I am not planning to include new features in Easy Random as the project is in maintenance mode.

Thank you for your PR anyway, much appreciated.

